PR #7, opened 2026-04-27
ci: add lean GitHub Actions workflow with hermetic auth env

timfee:
> ⚠️ **PR #8 is stacked on this branch.** Merging #7 first lets GitHub auto-redirect #8's base from `timfee-patch-2` → `main`. If this is squash-merged, #8 will need a manual rebase (`git rebase --onto main timfee-patch-2 chore/cleanup-npm-scripts`).

## Summary
- Replace the matrix-based template (Node 20 + 22) with **four parallel jobs** (`lint`, `test-unit`, `test-integration-fake`, `test-smoke`) on a single Node 20 + ubuntu-latest runner. Each failure surfaces as its own PR check, so a flaky smoke test does not bury a real unit-test regression.
- Add a workflow-level **hermetic auth env block** (`GOOGLE_APPLICATION_CREDENTIALS=/nonexistent`, `GCE_METADATA_HOST=127.0.0.1:1`, `NO_GCE_CHECK=true`). Any test that reaches `getAuthClient()` fails fast with a named ENOENT instead of timing out on metadata-server discovery.
- Fix `test/local/chromepolicy.test.js` to inject `{ auth: {} }` so `RealChromePolicyClient` short-circuits ADC via `apiOptions.auth` — required for `test:unit` to be green under the hermetic env.
- Add concurrency group keyed on PR number so old runs cancel when you push again. Minimal `permissions: contents: read`.

## Out of scope (follow-up)
- Script rationalization (`npm test` ≠ `presubmit`, `format` overlaps `lint:fix`, `test:full` looks orphaned, etc.) — separate PR (#8).
- Auditing other `test/local/*.test.js` files for ADC leaks. The hermetic env catches them with a clear failure; we fix as they surface.
- Branch protection rules (needs repo admin, not a code change).
- `test:integration:real` and `postsubmit` are intentionally not run by this workflow — they use real ADC and live elsewhere.

## Test plan
- [x] `GOOGLE_APPLICATION_CREDENTIALS=/nonexistent GCE_METADATA_HOST=127.0.0.1:1 NO_GCE_CHECK=true npm run test:unit` → 313/313 pass
- [x] Same env: `npm run lint` passes
- [x] Same env: `npm run test:integration:fake` → 11 pass / 3 skipped / 0 fail
- [x] Same env: `npm run test:smoke` exits 0
- [x] Local `npm run presubmit` passes
- [ ] CI run on this PR is green across all 4 jobs

PR #9, opened 2026-04-27
docs: restructure README, add docs/, fix stale claims

timfee:
## Summary

Bundled doc audit and restructuring. Independent of #7 and #8 (no overlap).

### Restructure
- Split the operations half of the root README into a new `docs/` directory: `configuration.md`, `troubleshooting.md`, `architecture.md`, `faq.md`, plus a `README.md` index.
- Trim root `README.md` from **635 → 325 lines**. The Quick Start (clone, install, auth, client config snippets, verify) stays fully intact in root, since that is the conversion path for viewers landing on the repo from the launch video.
- Add `lib/README.md` (was missing) as a five-line index of `lib/api/`, `lib/util/`, `lib/knowledge/`, and `lib/constants.js`.

### Stale claims fixed
- **`lib/api/README.md`**: removed the `fake_*_client.js` pattern it described — those files never existed. The actual pattern: instantiate the same `Real*Client` with `{ rootUrl, auth: fakeAuth }` to redirect at the in-process Express fake server (`test/helpers/fake-api-server.js`). Cross-checked against `mcp-server.js` and `test/helpers/integration/tools/client_factory.js`.
- **`lib/api/README.md` legacy-wrappers section**: previously claimed `chromepolicy.js` was unused. It actually still exports `ConnectorPolicyFilter`, which `tools/definitions/get_connector_policy.js` and `tools/definitions/diagnose_environment.js` import. Updated to call this out and note the prerequisite for deletion.
- **`tools/README.md` Tool Groups table**: removed the `cep_feedback` row (no such tool exists in the codebase), added `get_dlp_rule` to the DLP row, and added a Diagnostics row for `diagnose_environment` (which existed but was unlisted).
- **Root `README.md` Development**: corrected misleading "after merge, postsubmit adds a real-API run" to reflect that postsubmit runs the full presubmit suite *plus* the real-API integration pass.

### Content additions surfaced by the launch video
- `docs/configuration.md`: paragraph on the OAuth roadmap — today, OAuth requires a BYO client; a Google-managed OAuth client for CEP is coming.
- `docs/troubleshooting.md`: the `/mcp` reload tip when CEP tools do not show up after a client restart.

### Smaller fixes
- Trimmed eval-command duplication in `test/evals/README.md:266–298` to lead with the four npm scripts and reference the runner's `--help` block.
- Style pass on edited files: close-set em-dashes (`word—word`), colons inside boldface for `**Term:**` definition-list patterns, and semicolons where a single em-dash was joining two complete clauses.

## Out of scope (not in this PR)
- Stripping eval aliases (`eval:agentic`, `eval:full`, `eval:knowledge`) — both READMEs reference each one; kept all four.
- Per-tool documentation under `tools/` — `tools/README.md` covers the registration pattern; tool-specific docs are not needed.
- Knowledge-base content review (`lib/knowledge/*.md`) — those are domain content, not project docs.
- A separate `docs/development.md` — `CONTRIBUTING.md` and `test/README.md` already serve that role.
- Sweeping the existing Quick Start's prose style — left as-is to keep the diff focused.

## Test plan

- [x] `npx prettier --check .` passes on the full tree.
- [x] `npm run presubmit` passes locally (verified via husky pre-commit hook).
- [x] Every link in the new root `README.md` and new `docs/*.md` files resolves to an existing file.
- [x] All `lib/api/*` claims cross-checked against `mcp-server.js`, `client_factory.js`, and the actual file listing in `lib/api/`.
- [x] All `tools/README.md` table entries cross-checked against `tools/definitions/` directory contents and `tools/index.js` registrations.
- [ ] CI green on this PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,11 +27,11 @@ This project follows
 
 If you encounter an issue, please file a GitHub issue with the following:
 
-1. **Generate a diagnostic report** — in your Gemini CLI session, use the `/bug`
-   command. This creates a diagnostic file with session logs and environment
-   details. Attach it to the issue.
+1. **Generate a diagnostic report.** In your Gemini CLI session, use the
+   `/bug` command. This creates a diagnostic file with session logs and
+   environment details. Attach it to the issue.
 
-2. **Run presubmit** — run `npm run presubmit` and include the output. This
+2. **Run presubmit.** Run `npm run presubmit` and include the output. This
    helps determine whether the issue is environmental or a code bug.
 
 3. **Describe what you expected** vs. what actually happened, including the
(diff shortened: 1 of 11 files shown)


dylan-klein:
Is this a duplicate of #10 or intended to also be submitted? It seems like a duplicate because both PRs perform the same major restructuring (moving root README content into a new `docs/` directory) and fix identical stale claims in `lib/api/README.md` regarding the API client patterns.

dylan-klein:
Correction: I see now that #10 is stacked on this branch. Since #10 includes all the restructuring from this PR in its diff, it appeared to be a duplicate at first glance. My mistake.

dylan-klein (approved):
Restructure looks solid. Logic and links verified.

PR #13, opened 2026-04-28
chore: simplify DLP rule and posture review prompts

timfee:
## Summary

The `/cep:optimize` prompt and its supporting knowledge articles asked the agent to classify environments into a four-tier maturity model and to evaluate rules against a four-dimension framework. The agent surfaced those labels to administrators as if they were documented Chrome Enterprise Premium concepts.

This PR rewrites the optimize prompt and the two supporting articles to present the same evaluation logic in plain language, treating the heuristics as internal reasoning aids rather than user-facing taxonomy.

## Changes

- **`prompts/definitions/optimize.js`** — Output template now leads with an `Environment summary` paragraph and a `Rule findings` section per rule (`What we found / Why it matters / Recommended change / Patch`), followed by `Suggested next actions`. Removes the prior "Maturity Assessment" section and the per-rule "MECE dimension violation" line.
- **`prompts/system-prompt.md`** — Security guidance workflow drops the explicit "assess Tier" step. Adds a constraint forbidding the agent from naming internal frameworks, tier numbers, or model labels in replies; findings must use plain CEP nouns (licenses, connectors, SEB extension, AUDIT/WARN/BLOCK).
- **`lib/knowledge/12-security-posture-guide.md`** — Title changes to "CEP Security Posture Assessment Heuristics". Rewrites the body around four ordered questions (prerequisites in place? rules exist? rules tuned? rules enforcing?) with the same remediation steps under each.
- **`lib/knowledge/15-rule-quality-guidelines.md`** — Title changes to "DLP Rule Evaluation Heuristics". Drops the `<mece_api_heuristics_framework>` and `<dimension_N>` wrappers; presents the eight evaluation heuristics as a flat list. Frontmatter `summary` rewritten to reflect the new framing.
- **`test/local/prompts.test.js`** — Assertions updated for the new template; adds regression checks that the prompt template never contains "Maturity Tier", "MECE", or "Tier 0–3".
- **`test/evals/cases/prompts/optimize.md`** — Golden responses and judge instructions for `pr08`–`pr14` rewritten in plain language. Each judge instruction now treats internal-taxonomy leakage in the agent's reply as a soft FAIL.
- **`test/evals/scenarios/{no-dlp-rules,overly-broad-block-rule}.js`** — Comments updated to match the renamed eval tag.
- **`README.md`, `prompts/README.md`** — `cep:optimize` description updated to match the new prompt focus.

## Test plan

- [x] `npm run lint` passes.
- [x] `npm run test:unit` passes (321/321).
- [x] `npm run test:smoke` passes.
- [x] `node test/evals/run.js --category prompt --dry-run` passes (19/19 cases load).
- [x] `npm run eval` against a fake backend with a live Gemini judge — verify the agent's actual replies under the new system prompt no longer surface "Tier" / "MECE" / "framework" labels and the LLM judge grades them PASS.
- [x] Spot-check `/cep:optimize` and `/cep:health` interactively against the fake API server. Confirm output uses plain language ("the environment has 4 audit-only rules" rather than "Tier 1 (Visibility)") and that the agent does not cite "Maturity Tier" or "MECE Framework".

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,7 +233,7 @@ For environment variables and stdio vs. HTTP transport, see
 | Prompt         | Description                                                                            |
 | :------------- | :------------------------------------------------------------------------------------- |
 | `cep:health`   | Health check of the Chrome Enterprise environment (APIs, DLP, connectors, extensions). |
-| `cep:optimize` | Rule optimization and maturity assessment.                                             |
+| `cep:optimize` | Rule-by-rule review with tuning, enforcement, and cleanup recommendations.             |
 | `cep:expert`   | Manually re-injects the expert persona and rules (useful if the agent loses context).  |
 
 ### Tools
--- a/lib/knowledge/12-security-posture-guide.md
+++ b/lib/knowledge/12-security-posture-guide.md
@@ -1,53 +1,45 @@
 ---
-title: CEP Security Posture & Remediation Guide
+title: CEP Security Posture Assessment Heuristics
 articleId: '12'
-summary: 'Definitive roadmap for improving security posture using the CEP Maturity Model (Tiers 0-3). Helps administrators transition from visibility to active protection. Covers the critical "Telemetry Dependency" (logs require active rules) and remediation for each tier. Keywords: Maturity Tiers, Starter Pack rules, Telemetry Dependency, SEB extension requirements.'
+summary: 'Internal evaluation criteria the agent uses to assess a Chrome Enterprise Premium environment and recommend next steps. Walks through whether the prerequisites (licenses, connectors, SEB extension) are present, whether DLP rules exist, whether they are tuned, and whether they are enforcing. Covers the telemetry dependency (logs require active rules). For agent-internal use only — do not surface labels or framework names to users.'
 ---
 
-# CEP Security Posture & Remediation Guide
+# CEP Security Posture Assessment
 
-This guide is the primary technical reference for improving a Chrome Enterprise Premium (CEP) environment. To provide effective advice, first **Diagnose** the environment using `diagnose_environment` and **Assess** the current Maturity Tier based on the criteria below.
+These criteria help the agent decide what to recommend after running `diagnose_environment`. They are agent-internal — describe findings to the user in plain language without quoting category names from this article.
 
-## Maturity Tiers
+## Step 1: Are the prerequisites in place?
 
-### Tier 0: Foundation (Prerequisites)
-
-**Criteria**: Missing CEP licenses, unconfigured connectors, or missing Secure Enterprise Browser (SEB) extension.
-**Goal**: Establish the infrastructure for scanning and enforcement.
-**Remediation Steps**:
+Look for missing CEP licenses, unconfigured Content Analysis Connectors, or a missing Secure Enterprise Browser (SEB) extension. If any are missing, recommend:
 
 1.  **Licenses**: Assign CEP licenses to all managed users.
 2.  **Connectors**: Enable and configure all Content Analysis Connectors (Upload, Download, Paste, Print, URL Check) at the Root OU.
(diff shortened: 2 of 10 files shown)


timfee:
## Live eval run

`npm run eval -- --category prompt` against this branch with `CEP_BACKEND=fake` and a live Gemini judge: **15/19 prompt cases passed (78.9%)**.

| Case | Result | Diagnosis |
| :--- | :----- | :--- |
| `pr08`, `pr13` | FAIL | Agent returned an empty response with no tool calls — transient Gemini API behavior, re-runs typically pass. Not caused by this PR. |
| `pr12` | FAIL | Agent called the expected tools but the judge wanted log-driven analysis to lead, not rule-config heuristics. Borderline; tracked in a follow-up issue. |
| `pr16`, `pr17` | FAIL | Both are `cep:health` cases not touched by this PR. Judge wants synthesis ("the missing connectors render DLP non-functional") rather than per-control status rows. Pre-existing; tracked in a follow-up issue. |
| `pr08`–`pr11`, `pr14` | PASS | The `cep:optimize` cases this PR rewrote. Judge reasoning explicitly notes the agent avoids the previously leaked taxonomy ("avoiding prohibited internal taxonomy"). |

The optimize prompt rewrite holds up in live grading. Failures break down to model variance (`pr08`, `pr13`), one borderline judge call (`pr12`), and two pre-existing `cep:health` issues unrelated to this PR.

dylan-klein (commented):
In the judge instructions for pr08, pr09, pr10, and pr11, could you clarify how a 'soft FAIL' is meant to be interpreted by the LLM judge or the runner? Since the evaluation logic in `lib/judge.js` and `run.js` is strictly binary (PASS/FAIL), it is not clear how the judge will interpret it.

timfee:
## Eval comparison: directive strengthening (commit 4cb8f72 → 9bbfa56)

`npm run eval -- --category prompt` against the fake backend, two runs at each commit (LLM-judge variance is real, so single runs aren't conclusive).

| | Run 1 | Run 2 | Failures |
| :--- | :--- | :--- | :--- |
| **Before** (`4cb8f72`, no directives) | 13/19 (68%) | 15/19 (79%) | `pr08`, `pr09`, `pr10`, `pr11`, `pr12`, `pr13`, `pr16`, `pr17` (each ≤ 2 of 4 runs) |
| **After** (`9bbfa56`, with directives) | 14/19 (74%) | 15/19 (79%) | `pr08`, `pr09`, `pr10`, `pr12`, `pr13`, `pr16`, `pr17` (similar spread) |

Net: the directive-strengthening did not produce a measurable change against the per-run noise floor. Pass rates overlap; the same handful of cases drift between PASS and FAIL across runs regardless of which version is on disk. With four runs and 19 cases per run we don't have enough samples to detect a small effect.

The strengthened directives stay anyway because:
1. They give the agent an explicit ordering it can fall back on (analyze logs first, prioritize tuning audit-only rules, stage new BLOCK rules in AUDIT).
2. They make the prompt more readable for human contributors and replace the deleted Tier-keyed prescriptions like-for-like.
3. The cases consistently failing across both versions (`pr12`, `pr16`, `pr17`) are not addressable from this article — they're tracked in #15 (`pr12` log-first ordering, fixed in #18) and #14 (`pr16`, `pr17` synthesis, fixed in #17).

## Dylan's review feedback

Pushed `35e1733` to address [@dylan's review](https://github.com/google/chrome-enterprise-premium-mcp/pull/13#pullrequestreview-4193005755). The four `pr08`–`pr11` rubrics that used "soft FAIL" are rewritten as a hard FAIL with explicit triggering language: _"If the agent uses internal taxonomy in the user-facing reply (e.g., 'Tier 0', 'MECE'), grade as FAIL even if the rest of the analysis is correct."_ Same intent (taxonomy leakage disqualifies a response), now expressible against the strict PASS/FAIL output the judge in `lib/judge.js` produces.

dylan-klein (approved):

PR #37, opened 2026-04-29
fix(tools): don't claim ENABLED before the long-running enable completes

### LINKED ISSUE #27: check_and_enable_cep_api claims success without polling the long-running operation
`enableService` returns a Service Usage long-running operation (often `{ done: false, name: 'operations/...' }` for async enables), but the handler awaits the call and immediately marks the API as ENABLED:

https://github.com/google/chrome-enterprise-premium-mcp/blob/a608ee6d6d3bf3d6095ffc4afd9531bacdd1b44e/tools/definitions/check_and_enable_cep_api.js#L86-L95

There is no check of the `done` field and no polling. For APIs that take time to enable, the summary tells the user the API is enabled before it actually is, leading to follow-up tool calls that fail because the API is still propagating.

Fix: poll the operation until `done === true` (with a sane timeout), or make the success message clearer that the enable is in-flight ("requested" vs "enabled").

timfee:
Closes #27.

`enableService` returns a long-running operation. The handler awaited the call and immediately reported ENABLED — a false positive when the LRO was still pending. The new behavior: if `response.done` is true, mark ENABLED (as before); otherwise mark ENABLING and surface the operation name with "may take a few minutes".

Verified locally: unit tests pass; integration tests pass on the fake backend. Pre-existing port-binding failures in `mcp-server.test.js` are unrelated.


dylan-klein:
We're returning `operationName` in the structured data, but since there isn't currently a tool to check operation status, this is a "dead end" for an agent. Perhaps we should instead add an explicit instruction to the natural language summary (e.g., `...enable requested, may take a few minutes. Re-run this tool to verify status.`). Do we want to keep the operation ID in the structured data if the agent can't act on it?

timfee:
Good catch. Splitting the difference — agreed the summary text shouldn't carry an ID the agent can't act on, but I'd keep the operation name in `structuredContent` (it's forensic for humans reading agent traces, links to the GCP Console operations page, and a future `get_operation_status` tool could consume it).

Pushed 47a790d with three changes:

1. Dropped `(operation: ...)` from the summary line.
2. Appended `Re-run this tool to verify status.` to the summary, since `check_and_enable_cep_api` already starts with `getServiceStatus` so re-running is the verification path.
3. Only includes `operationName` in `structuredContent` when the upstream response actually carried one — the previous fallback could print `(operation: \`unknown\`)` to the user, which was strictly worse than nothing.

LMK if you'd rather drop `operationName` from `structuredContent` entirely; happy to take that further.

tools/definitions/check_and_enable_cep_api.js
```diff
@@ -87,9 +87,20 @@ This is a PREREQUISITE tool. Many other tools will fail if necessary APIs are di
...
-                await serviceUsageClient.enableService(projectId, api, authToken)
-                results.push(`- **${api}** — NEWLY_ENABLED (project: \`${projectId}\`)`)
-                apiStatuses.push({ apiName: api, status: 'ENABLED', projectId })
+                const enableResponse = await serviceUsageClient.enableService(projectId, api, authToken)
+                if (enableResponse && enableResponse.done === true) {
+                  results.push(`- **${api}** — NEWLY_ENABLED (project: \`${projectId}\`)`)
+                  apiStatuses.push({ apiName: api, status: 'ENABLED', projectId })
+                } else {
+                  results.push(
+                    `- **${api}** — ENABLING (project: \`${projectId}\`): enable requested, may take a few minutes. Re-run this tool to verify status.`,
```
dylan-klein, line 103: This "else" block currently acts as a catch-all for any response that isn't a confirmed success. If the API call returns an unexpected result or a silent error (like a billing issue or a lack of permissions) without throwing an exception, the tool will incorrectly report the status as "ENABLING" and instruct the user to "wait a few minutes."

This leads to a poor experience where a user or an AI agent stays in a loop, waiting for an operation that will never complete because it actually failed. To prevent this, the code should explicitly check for the pending state and handle unknown or malformed responses as an error.

```javascript
// Current implementation
if (enableResponse && enableResponse.done === true) {
  // reports SUCCESS
} else {
  // reports ENABLING for everything else, including errors
  results.push(`- **${api}** — ENABLING...`)
}

// Suggested implementation
if (enableResponse && enableResponse.done === true) {
  results.push(`- **${api}** — NEWLY_ENABLED...`)
} else if (enableResponse && enableResponse.done === false) {
  // Only report ENABLING if the API explicitly says it is pending
  results.push(`- **${api}** — ENABLING...`)
} else {
  // Handle unexpected or null responses as an error immediately
  results.push(`- **${api}** — ERROR: Received unexpected response. Please check project settings.`)
}
```

dylan-klein (commented):
Technical review of LRO handling

tools/definitions/check_and_enable_cep_api.js
timfee: Agree with the diagnosis; pushed 93e3ce6 with a slightly extended fix.

Your three-way split (`done:true` → success, `done:false` → ENABLING, else → error) catches the malformed-response case. It also surfaces a related bug worth handling at the same time: Google's LRO contract sets `error` when `done: true` and the operation completed unsuccessfully. The previous code's `if (done === true) → NEWLY_ENABLED` would have reported billing or permission failures as success — strictly worse than reporting them as still pending.

Final ordering, error-first:

1. `enableResponse?.error` → FAILED (with the upstream error message). Catches `done: true, error: {...}` LROs.
2. `enableResponse?.done === true` → NEWLY_ENABLED.
3. `enableResponse?.done === false` → ENABLING (with the existing "Re-run this tool to verify status" hint).
4. anything else → UNKNOWN with a re-run hint, so the caller does not loop on a non-pending response.

Two new unit tests cover the LRO-error and unknown-shape branches. `node --test test/local/service_usage.test.js` is 14/14 green; `npm run lint` clean.

LMK if you'd rather hold the FAILED branch for a follow-up — the change addresses your point on its own and the LRO-error case can be a separate commit if you'd prefer.

dylan-klein (approved):
LGTM. The revised logic completely addresses the 'zombie state' concern and the addition of the FAILED branch is a great improvement. No need to split it out — it makes sense to include it here.

Issue #20, opened 2026-04-29
get_connector_policy returns data shape that doesn't match its declared outputSchema


timfee:
Three layered mismatches between `get_connector_policy`'s declared schema and what the tool actually returns:

1. **`data` is a bare array but the schema declares an object.** The handler passes `formattedPolicies` (an array) as `data`, so `content[1]` (the JSON code block) is a top-level array. The `outputSchema` declares `{ connectorPolicies: [...], connectorType: string, orgUnitId: string }`.
2. **`structuredContent` adds an undeclared `configured` field** not in the schema.
3. **Item shape is wrong.** Schema says each item is `resolvedChromePolicy.extend({ isEnabled })` (i.e. `{ targetKey, value, isEnabled, ... }`). Actual items are flat human-readable dicts like `{ "serviceProvider (describe to user as 'Provider')": "Chrome Enterprise Premium", isEnabled: true }`. The `targetKey`/`value` fields are absent.

https://github.com/google/chrome-enterprise-premium-mcp/blob/a608ee6d6d3bf3d6095ffc4afd9531bacdd1b44e/tools/definitions/get_connector_policy.js#L317-L326

Result: SDK clients reading `structuredContent` see one shape; clients reading the JSON in `content[1]` see another; neither matches what the schema advertises. Either fix the schema to document the actual shape, or restructure the response to match.

timfee:
On re-review the runtime impact is smaller than I originally framed: the `outputSchema` uses Zod's `.passthrough()`, so unexpected fields aren't rejected and the SDK doesn't fail validation. The actual issue is that the schema is misleading documentation — it advertises `{ targetKey, value, isEnabled }` but every item in `connectorPolicies` is a flat human-readable dict like `{ "serviceProvider (describe to user as 'Provider')": "...", isEnabled: true }`. The `data` array vs `structuredContent` object mismatch and the undeclared `configured` field are also real, but no agent or client will silently break because of any of these.

Better treated as a P1 documentation/contract cleanup than a P0 bug. Either rewrite the schema to describe the actual shape, or restructure the response to match the existing schema.

PR #40, opened 2026-04-29
[abandoned] fix(tools): align get_connector_policy outputSchema with actual response shape

timfee:
[abandoned]

Originally opened to fix the `get_connector_policy` outputSchema, which listed items as `resolvedChromePolicy.extend(...)` (the wrong shape for the actual response), and to wrap the `data` argument in a matching envelope so `content[1]` and `structuredContent` agreed. Addressed #20.

Abandoned because the fix was not landed in this form.


timfee:
Superseded by #102, which applies the same fix plus a shared `payload` variable so `data` and `structuredContent` cannot drift, and `.describe()` strings on the new schema fields. Closing this in favor of the consolidated PR.

PR #45, opened 2026-04-29
fix(server): scope session state per HTTP request

### LINKED ISSUE #19: Shared session state leaks customerId across users in HTTP/SSE mode
`mcp-server.js` constructs a single `sharedSessionState` object once at startup and passes it to every `/mcp` POST and `/sse` GET handler:

https://github.com/google/chrome-enterprise-premium-mcp/blob/a608ee6d6d3bf3d6095ffc4afd9531bacdd1b44e/mcp-server.js#L295-L320

`tools/utils/wrapper.js` writes the resolved `customerId` and `cachedRootOrgUnitId` into that object on first use:

https://github.com/google/chrome-enterprise-premium-mcp/blob/a608ee6d6d3bf3d6095ffc4afd9531bacdd1b44e/tools/utils/wrapper.js#L192-L207

In stdio mode (one process per user) this is fine. In HTTP/SSE mode, two concurrent users from different Workspace tenants will share the same cache: the first to resolve writes their customerId, and every subsequent tool call uses that ID regardless of whose auth token is on the request. Subsequent tool calls return data from the wrong tenant.

Fix: scope session state per-request (or per OAuth principal) in HTTP mode rather than process-wide.

timfee:
Closes #19.

In stdio mode the process-wide `sharedSessionState` is correct (one user per process). In HTTP/SSE mode `sharedSessionState` was passed to every request handler, so concurrent users from different Workspace tenants shared the same `customerId` and orgUnit cache. Construct a fresh `sessionState` inside each `/mcp` POST and `/sse` GET handler. New unit test documents the per-request contract.

Verified locally: unit tests pass; integration tests pass on the fake backend. Pre-existing port-binding failures in `mcp-server.test.js` are unrelated.


test/unit/mcp-server-state-isolation.test.js
```diff
@@ -0,0 +1,97 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * @file Verifies that each HTTP request receives a distinct sessionState object
+ * so that resolved customer data from one tenant cannot bleed into another.
+ */
+
+import assert from 'node:assert/strict'
```
shantand1234, line 25: this unit test is not testing anything specific. Its just testing general object properties

test/unit/mcp-server-state-isolation.test.js
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Simulates the per-request sessionState factory used inside each HTTP handler.
+ * This mirrors the inline object construction in mcp-server.js so that the test
+ * stays decoupled from Express and MCP SDK internals.
+ */
+function makeSessionState() {
+  return { customerId: null, cachedRootOrgUnitId: null, pendingRule: null, history: [] }
+}
+
+describe('HTTP per-request session state isolation', () => {
```
shantand1234: this test is not actually testing the http per request

timfee:
@shantand1234 — thanks for the earlier review. Pushed c32d4af which addresses both comments. Summary of what changed and why it now tests the per-request invariant:

**Refactor (mcp-server.js):**
- Extracted three named exports — `createSessionState()`, `createMcpPostHandler(gcpInfo, getServerImpl?)`, `createSseHandler(gcpInfo, sseTransports, getServerImpl?)`. The handler factories accept an optional `getServerImpl` so tests can inject a recorder; default is the real `getServer`.
- Replaced the inline `app.post('/mcp', async (req, res) => { ... })` and `app.get('/sse', async (_req, res) => { ... })` blocks with calls to those factories. No behavior change in production — the closures that ran before are now the same closures returned by the factories.
- Gated `main()` behind an entry-point check (`if (import.meta.url === \`file://${process.argv[1]}\`)`) so importing the module from a test file doesn't auto-start the server.

**Test (test/unit/mcp-server-state-isolation.test.js):**
- Builds a recorder for `getServer` that pushes its `sessionState` argument into an array.
- Invokes the actual `createMcpPostHandler` (and `createSseHandler`) handler twice with stubbed `req`/`res`.
- Asserts `captured[0] !== captured[1]` (distinct references, which is the regression the PR is fixing) and additionally mutates `captured[0].customerId = 'C0TENANT1'` then asserts `captured[1].customerId === null` — that's the actual cross-tenant cache invariant in plain words.

So the test now exercises the production handler code path (not a local mirror), and the assertions describe the invariant the fix is enforcing rather than just the shape of the object.

Mind taking another look?

shantand1234 (approved):

PR #68, opened 2026-04-29
fix(evals): default EXPERIMENT_DELETE_TOOL_ENABLED=true so m03 tests real agent behavior

### LINKED ISSUE #51: Eval m03 (delete_agent_dlp_rule) semantics flip based on EXPERIMENT_DELETE_TOOL_ENABLED
test/evals/cases/tools/delete_agent_dlp_rule.md case m03 expects the agent to refuse deletion (`expected_tools: []`). But `delete_agent_dlp_rule` only registers when EXPERIMENT_DELETE_TOOL_ENABLED is set in the environment.

The eval runner doesn't set the flag → the tool isn't registered → the agent 'passes' by being unable to call it. Run with the flag set → the agent might call the tool and 'pass' by deleting it. Same case, opposite behaviors, no signal in the result.

Fix: pin the flag in the eval runner (so the test reflects the production-with-experiment behavior), or split into two cases (one for each flag state), or skip the case if the flag isn't set with a clear message.

(Split from #31)

timfee:
m03 expects the agent to refuse deletion. But delete_agent_dlp_rule only registers when EXPERIMENT_DELETE_TOOL_ENABLED is set — without the flag the tool isn't there and m03 'passes' for the wrong reason. Defaults the flag in the eval runner so m03 actually exercises agent judgment; user can still override.

Closes #51.

### DIFF
--- a/test/evals/README.md
+++ b/test/evals/README.md
@@ -288,13 +288,14 @@ The runner accepts more flags (`--runs`, `--concurrency`, `--delay`,
 
 ### Environment variables
 
-| Variable         | Description                                                    | Default |
-| ---------------- | -------------------------------------------------------------- | ------- |
-| `GEMINI_API_KEY` | **Required.** Gemini API key for both the agent and the judge. | --      |
-| `CEP_BACKEND`    | `fake` (in-process mock) or `real` (live Google APIs).         | `fake`  |
-| `EVAL_CATEGORY`  | Alternative to `--category` flag.                              | --      |
-| `EVAL_IDS`       | Alternative to `--id` flag. Comma-separated.                   | --      |
-| `EVAL_TAGS`      | Alternative to `--tags` flag. Comma-separated.                 | --      |
+| Variable                        | Description                                                                                                                   | Default |
+| ------------------------------- | ----------------------------------------------------------------------------------------------------------------------------- | ------- |
+| `GEMINI_API_KEY`                | **Required.** Gemini API key for both the agent and the judge.                                                                | --      |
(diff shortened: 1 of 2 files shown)


test/evals/README.md
```diff
@@ -288,13 +288,14 @@ The runner accepts more flags (`--runs`, `--concurrency`, `--delay`,
...
+| `GEMINI_API_KEY`                | **Required.** Gemini API key for both the agent and the judge.                                                                | --      |
+| `CEP_BACKEND`                   | `fake` (in-process mock) or `real` (live Google APIs).                                                                        | `fake`  |
+| `EVAL_CATEGORY`                 | Alternative to `--category` flag.                                                                                             | --      |
+| `EVAL_IDS`                      | Alternative to `--id` flag. Comma-separated.                                                                                  | --      |
+| `EVAL_TAGS`                     | Alternative to `--tags` flag. Comma-separated.                                                                                | --      |
+| `EXPERIMENT_DELETE_TOOL_ENABLED`| Registers the delete-tool experiment. The runner defaults this to `true` so cases like `m03` test real agent judgment. Set to `false` to disable. | `true`  |
```
shantand1234, line 298: What is m03 ?

shantand1234 (approved):

test/evals/README.md
timfee: https://github.com/google/chrome-enterprise-premium-mcp/blob/main/test/evals/cases/tools/delete_agent_dlp_rule.md

PR #72, opened 2026-04-29
fix(util): make gcloud quota-project probe async with per-call timeout

### LINKED ISSUE #55: auth-error.js issues up to 12 sequential execFileSync gcloud calls on the event loop
lib/util/auth-error.js suggestQuotaProject runs synchronously: one `gcloud config get-value`, one `gcloud projects list`, then up to 10 `gcloud services list --project=...` checks, all via execFileSync.

This is called from getAuthErrorMessage which is called from callWithRetry which runs in every API method's error handling. On a machine with many GCP projects and a slow gcloud CLI, this blocks the JS event loop for 10–50 seconds during error handling. Move to async `execFile` with a per-call timeout, or move the project suggestion to a background worker.

https://github.com/google/chrome-enterprise-premium-mcp/blob/3c7a9629f68d0f358d9e13c666a8d3a74b3689ca/lib/util/auth-error.js#L70-L115

(Split from #32)

timfee:
Closes #55.

`suggestQuotaProject` in `lib/util/auth-error.js` issued up to 12 sequential `execFileSync('gcloud', ...)` calls on the event loop. The function is reachable from `getAuthErrorMessage` → `callWithRetry`, which runs in every API method's error handler. On a workstation with many GCP projects, the event loop was blocked for 10–50 seconds during error handling.

This PR converts the calls to async `execFile` with a 1s per-call timeout and a 5s total budget. On timeout or error the function continues with what it has collected. `getAuthErrorMessage` is now async; call sites in `helpers.js` are awaited. Existing `auth.test.js` mocks are updated to the callback-style `execFile`.

### DIFF
--- a/lib/util/auth-error.js
+++ b/lib/util/auth-error.js
@@ -21,99 +21,113 @@ limitations under the License.
  * quota project) and returns actionable remediation instructions.
  */
 
-import { execFileSync } from 'node:child_process'
+import { execFile } from 'node:child_process'
 import { ERROR_MESSAGES, SCOPES } from '../constants.js'
 
-let cachedIsGcloudInstalled = null
+const GCLOUD_CALL_TIMEOUT_MS = 1000
+const GCLOUD_TOTAL_BUDGET_MS = 5000
+
+let cachedIsGcloudInstalled = /** @type {boolean|null} */ (null)
+let gcloudCheckPromise = /** @type {Promise<boolean>|null} */ (null)
+
+/**
+ * Runs a gcloud command with a per-call timeout. Returns stdout or null on error/timeout.
+ * @param {string[]} args - Arguments to pass to gcloud.
+ * @returns {Promise<string|null>} The stdout output, or null if the call fails or times out.
+ */
+function runGcloud(args) {
+  return new Promise(resolve => {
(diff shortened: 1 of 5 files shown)


dylan-klein (commented):
There appear to be two critical call sites in `lib/util/auth.js` and `lib/util/google-auth-provider.js` that are missing the `await` keyword. Since `getAuthErrorMessage` is now async, these will currently throw a cryptic `Error: [object Promise]` instead of the gcloud instructions.

As an optional improvement, a unit test could be added to `test/local/auth.test.js` that specifically triggers an error in `getAuthClient` (by mocking a failure in `getClient()`). This would verify the error path and prevent future regressions of this async behavior.

timfee:
Right call, both confirmed:

- `lib/util/auth.js:49` and `lib/util/google-auth-provider.js:49` were missing `await`. Fixed in fead54a.

There's a third site I introduced in PR #119 (the OAuth-cache fallback in `lib/util/auth.js#getAuthClient`). I went ahead and added `await` there too in PR #119 so whichever order these merge, no one inherits the `[object Promise]` bug.

Re: the unit test for the error path — I'll add it. Mocking `getClient()` to throw and asserting the message contains a remediation hint covers both sites and prevents this exact regression next time.

dylan-klein (approved):

Issue #22, opened 2026-04-29
list_customer_profiles silently swallows auth errors instead of triggering remediation


timfee:
The handler wraps the entire body in a try/catch and returns a success-shaped `formatToolResponse` with `error: true` buried inside `data` (no `isError: true` on the top-level response):

https://github.com/google/chrome-enterprise-premium-mcp/blob/a608ee6d6d3bf3d6095ffc4afd9531bacdd1b44e/tools/definitions/list_customer_profiles.js#L65-L115

Auth errors (401/403/PERMISSION_DENIED) never propagate to `guardedToolCall`, so the auth-remediation message that other tools surface for users (re-auth instructions, scope hints) never fires. The MCP client treats this as a successful tool call with an error string in the summary, and the user sees a generic "Error listing customer profiles" instead of an actionable next step.

No other tool wraps the entire handler this way — they let errors propagate to `guardedToolCall`, which has the centralized auth-error detection. This handler should follow that pattern (re-throw, or only catch specific non-fatal cases).

Issue #21, opened 2026-04-29
create_chrome_dlp_rule and create_default_dlp_rules unwrap result.response — may be wrong shape against real API


timfee:
`create_chrome_dlp_rule` extracts the created policy as `result.response`, with an inline comment acknowledging this may not survive contact with the real API:

https://github.com/google/chrome-enterprise-premium-mcp/blob/a608ee6d6d3bf3d6095ffc4afd9531bacdd1b44e/tools/definitions/create_chrome_dlp_rule.js#L262-L275

`RealCloudIdentityClient.createDlpRule` returns `response.data` from `policies.create`:

https://github.com/google/chrome-enterprise-premium-mcp/blob/a608ee6d6d3bf3d6095ffc4afd9531bacdd1b44e/lib/api/real_cloud_identity_client.js#L114-L139

The Cloud Identity `policies.create` endpoint is documented to return a long-running operation. Whether `response.data` is shaped `{ name, done, response }` or the policy directly depends on whether the operation completes synchronously. `create_default_dlp_rules` has the same unwrap pattern and then calls `createdPolicy.name` on the result — if `result.response` is undefined in production, this throws `TypeError` after the rule is created on the API, returning an error response despite a successful write.

Action: trace one real (non-fake) `policies.create` call and confirm the response shape, then either remove the `.response` unwrap or add a fallback. The existing comment is a known-unknown that should be resolved.

timfee:
Closing as a false positive on second look. Verified against the real Cloud Identity v1beta1 API on a test domain — `policies.create` returns exactly the same shape the fake server emits:

```json
{
  "done": true,
  "response": {
    "@type": "type.googleapis.com/google.apps.cloudidentity.policies.v1main.Policy",
    "name": "policies/akajj264aop7bpwicq",
    "customer": "customers/...",
    "setting": { ... }
  }
}
```

So `result.response` is populated synchronously, and `createdPolicy.name` works as written. The TODO comment in `create_chrome_dlp_rule.js:261-265` is alarmist and can be removed — but the runtime behavior is correct. Sorry for the noise.

timfee:
Reopening: re-verified that the surrounding TODO comment (`create_chrome_dlp_rule.js:261-265`) is misleading and should be removed. Verified `policies.create` against the real Cloud Identity v1beta1 API on a test domain — it returns:

```json
{ "done": true, "response": { "@type": "...Policy", "name": "policies/...", "setting": {...} } }
```

[...]

PR #16, opened 2026-04-28
Expand evaluation suite for get_connector_policy tool

TusharKanekiDey:
### Description
This PR significantly expands the evaluation suite and unit test coverage for the `get_connector_policy` tool. It ensures that the tool's evaluation logic is thoroughly audited across a wide range of configuration states for both Reporting and Non-Reporting (Analysis) connectors.

### Key Changes
1. **Consolidated & Granular Unit Tests:** Updated `test/unit/tools/get_connector_policy.test.js` with **36 tests** covering all requested scenarios (opt-in structures, missing events, granular analysis restrictions, and Print array handling).
2. **Deterministic Evaluation Suite:**
   - Created **11 dedicated backend scenario files** in `test/evals/scenarios/` to simulate specific connector states.
   - Rewrote `test/evals/cases/tools/get_connector_policy.md` into **atomic test cases** using these scenarios with consistent `tool__when__then` IDs.
   - Added `Judge Instructions` and `required_patterns` to ensure high-fidelity grading.
   - Added `connectors` to the documented evaluation categories in `README.md` and updated the `eval:agentic` script in `package.json` to include them.
   - Fixed `expected_tools` outliers in evaluation cases for consistency.

### Test Results
- **Unit Tests:** 100% Pass (349 total).
- **Evaluation Suite (Connectors Category):** 100% Pass (12/12 cases).
  - Verified: Missing connectors (with fix offers), Default vs. Customized events, Opt-in events, Delay enforcement warnings, Restricted scanning (allowlists), and 3rd-party provider bypasses.

PR #75, opened 2026-04-29
fix(security): address CodeQL alerts across regex/URL/proto handling

timfee:
Addresses 22 of the 24 currently-open CodeQL alerts on `main`. The remaining 2 (lib/constants.js #11, #12) are CodeQL data-flow false positives — the strings are scope URLs that get used as regex only inside one specific test interpolation; the constants themselves are correct.

## Production code

- **`tools/definitions/check_and_enable_cep_api.js`** (#19, #20) — replace `errorMessage.includes('serviceusage.googleapis.com')` with reliance on the existing `'Service Usage API'` wording branch (Google's canonical phrasing). Substring URL matching could match attacker-controlled lookalike subdomains; the wording branch is unambiguous and the catch flow is simpler.
- **`tools/definitions/knowledge.js`** (#1, #2, #3, #4) — harden the HTML cleaner:
  - Closing-tag patterns now allow optional whitespace inside (`</script\\s*>`), so well-formed HTML with whitespace in tags isn't bypassed.
  - A second pass strips orphan `<script` / `<style` opening tags whose closer is missing or malformed, so the substring can't survive.
  - Entity decoding switched from sequential `.replace` calls to a single regex pass keyed on a static map, so `&amp;lt;` decodes to `&lt;` (correct) instead of `<` (double-unescape bug).

## Test infrastructure

- **`test/helpers/fake-api-server.js`** (#5–#10, #21, #22) — introduce `isSafeKey()` and reject `__proto__` / `constructor` / `prototype` keys before assigning them to plain-object maps in the connector-policy batchModify endpoint and the licensing fixture loader. Coerce `req.query.filter` to a string before `.includes()` so an array-form filter (`?filter=a&filter=b`) can't trigger Array#includes semantics.

## Test files

- **regex hostname escaping** (#13, #14, #15) — escape `.` in hostname regexes in `test/local/auth.test.js`, `admin_sdk.test.js`, `check_cep_subscription.test.js`.
- **URL substring → anchored regex** (#16, #17, #18) — switch `text.includes('admin.googleapis.com')` style assertions to `/\\badmin\\.googleapis\\.com\\b/` matches in `utils.test.js` and `auth_messages.test.js`.

## Verification

- `npm run lint` clean
- `npm run test:unit` — 322/322 pass
- `npm run test:integration:fake` — 11/11 pass (3 skipped intentionally)

### DIFF
--- a/tools/definitions/check_and_enable_cep_api.js
+++ b/tools/definitions/check_and_enable_cep_api.js
@@ -107,10 +107,8 @@ This is a PREREQUISITE tool. Many other tools will fail if necessary APIs are di
                 errorMessage.includes('PERMISSION_DENIED') ||
                 errorMessage.includes('invalid_grant')
 
-              // Match Google's canonical wording rather than substring-matching
-              // the host (which CodeQL flags as incomplete URL sanitization and
-              // which would also match attacker-controlled lookalike subdomains).
-              const mentionsServiceUsage = errorMessage.includes('Service Usage API')
+              const mentionsServiceUsage =
+                errorMessage.includes('Service Usage API') || /\bserviceusage\.googleapis\.com\b/.test(errorMessage)
 
               if (isAuthError && !mentionsServiceUsage) {
                 throw error
--- a/tools/definitions/check_and_enable_cep_api.js
+++ b/tools/definitions/check_and_enable_cep_api.js
@@ -107,18 +107,16 @@ This is a PREREQUISITE tool. Many other tools will fail if necessary APIs are di
                 errorMessage.includes('PERMISSION_DENIED') ||
                 errorMessage.includes('invalid_grant')
 
-              if (
-                isAuthError &&
-                !errorMessage.includes('serviceusage.googleapis.com') &&
-                !errorMessage.includes('Service Usage API')
-              ) {
+              // Match Google's canonical wording rather than substring-matching
+              // the host (which CodeQL flags as incomplete URL sanitization and
+              // which would also match attacker-controlled lookalike subdomains).
+              const mentionsServiceUsage = errorMessage.includes('Service Usage API')
+
+              if (isAuthError && !mentionsServiceUsage) {
                 throw error
               }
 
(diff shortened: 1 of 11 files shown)


shantand1234 (commented):
Please fix the code ql alerts

shantand1234 (approved):

tools/definitions/check_and_enable_cep_api.js
```diff
@@ -107,18 +107,16 @@ This is a PREREQUISITE tool. Many other tools will fail if necessary APIs are di
...
+              // which would also match attacker-controlled lookalike subdomains).
+              const mentionsServiceUsage = errorMessage.includes('Service Usage API')
+
+              if (isAuthError && !mentionsServiceUsage) {
                 throw error
               }
 
-              const isServiceUsageError =
-                error.status === 403 ||
-                errorMessage.includes('serviceusage.googleapis.com') ||
```
shantand1234, line 120: isnt this a better check as its restricted to google apis , there can be a service usage api for something else too ? 

@TusharKanekiDey please review this change

timfee: Good point. `serviceusage.googleapis.com` is unambiguously Google's, while "Service Usage API" is generic enough that another vendor could use the same wording. The follow-up commit restores the host check alongside the prose phrase, using `/\bserviceusage\.googleapis\.com\b/`.

The CodeQL alert this addresses (`js/incomplete-url-substring-sanitization`) is about gating-on-host security decisions, where an attacker crafts a URL like `evil.com/serviceusage.googleapis.com/...` and a substring match treats it as the real host. Our usage isn't security-gating; we're classifying the body of an error returned by Google's API client to decide which user-facing message to show. The error string originates from Google's library, not from user input, so the spoofing scenario isn't reachable here. If CodeQL still flags the anchored shape on this commit, the right call is an inline suppression citing the threat-model mismatch, not stricter URL parsing.

PR #81, opened 2026-04-29
chore(husky): run lint-staged only in pre-commit; move test suite to pre-push

### LINKED ISSUE #78: Speed up the husky pre-commit hook
Sub-issue of #77.

The husky pre-commit hook currently runs the full `npm run presubmit` chain — `prettier --check . && npm run lint && npm run test:unit && npm run test:integration:fake && npm run test:smoke`. End-to-end that's around 90 seconds per commit on contributor hardware. Three observed consequences:

- Iteration friction. Anything that can be split into multiple commits often isn't, because the cost is paid per commit. Contributors who hit the wait often enough start reaching for `--no-verify`.
- Parallel commits queue on the hook, so coordinated work (multiple worktrees, automation, batch fixes) serializes on a check that CI is going to run again anyway.
- `prettier --check .` walks the entire working tree, including local-only state under `.claude/worktrees/` that isn't part of any commit. The hook then fails for unrelated reasons and forces a workaround (move the directory aside, retry).

The CI workflow at `.github/workflows/node.js.yml` already runs lint, unit, integration, and smoke as separate jobs on every PR. The duplication doesn't catch anything CI wouldn't catch first, with the exception of catching it before the contributor has pushed.

## Proposed approach

- Replace the pre-commit body with `npx lint-staged` so only files actually being committed are formatted and linted. Target: under five seconds.
- Move the heavy local check (`npm run test:unit && npm run test:smoke`) to a new `.husky/pre-push` hook. Skip integration there; CI runs it and the local feedback loop doesn't need it twice.
- Add `.claude/` to `.prettierignore` so worktree state stops fighting the hook.
- Enable `--cache` on prettier and eslint to make incremental runs effectively instant.

## Acceptance

- Pre-commit on a single-file change finishes in under 10 seconds on a typical laptop.
- Pre-push runs the unit suite + smoke before the push leaves the laptop.
- Existing CI gates remain unchanged so the safety net stays where it is.

timfee:
Closes #78.

The pre-commit hook ran the full presubmit chain on every commit (`prettier --check . && npm run lint && npm run test:unit && npm run test:integration:fake && npm run test:smoke`). The chain takes ~90 seconds. The same checks run as separate jobs in CI on every PR (`.github/workflows/node.js.yml`), so the hook duplicated CI.

Three changes:

1. `.husky/pre-commit` runs `npx lint-staged` only. Prettier and `eslint --fix` run on staged files only, with caching.
2. New `.husky/pre-push` runs `npm run test:unit && npm run test:smoke`. Integration tests stay in CI; they were already there.
3. `.prettierignore` excludes `.claude/` and `.gemini/`. Tool-local directories, not committed, but tripping the prettier check.

`lint` and `format` scripts pass `--cache`. CI gates are unchanged.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,0 +1,1 @@
+npm run test:unit && npm run test:smoke
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,1 +1,1 @@
-npm run test:unit && npm run test:smoke
+npm run test:unit && npm run test:smoke && npm run test:integration:fake
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,3 +1,3 @@
 # Run lint-staged to only apply autofixes to files that are actually being committed.
 # This prevents accidentally staging all tracked changes when only doing a partial commit.
-npx lint-staged && npm run presubmit
+npx lint-staged
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,3 +2,4 @@ test/evals/cases/**/*.md
 results.json
 results_final.json
 .claude/
+.gemini/
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,4 @@
 test/evals/cases/**/*.md
 results.json
 results_final.json
+.claude/


.husky/pre-push
```diff
@@ -0,0 +1 @@
+npm run test:unit && npm run test:smoke
```
shantand1234: should this be pre submit

timfee: Moved to pre-push because pre-submit was taking 90+ seconds every time and seemed overkill for submits (that should mostly handle static checking that is quick; e..g where we'd add typescript checking when we move to TS)

dylan-klein (commented):
Was removing the fake integration tests from the local workflow intentional? They take only ~1.7s and verify the full tool-to-API chain. I'd suggest adding `npm run test:integration:fake` to the new `.husky/pre-push` hook to catch regressions before they hit CI.

timfee:
Good call. Adding it.

Quick note on timing: `test:integration:fake` is ~8s here (14 tests, 11 pass, 3 skipped under `SKIP_SLOW=true`), not ~1.7s. Still cheap relative to the ~112s `test:unit + test:smoke` pre-push budget, so the trade is fine.

Updated `.husky/pre-push` to:

```bash
npm run test:unit && npm run test:smoke && npm run test:integration:fake
```

dylan-klein (approved):

PR #105, opened 2026-04-30
[abandoned] Add audit:scopes script for scope verification against live discovery docs

timfee:
[abandoned]

Originally opened to add `scripts/audit-scopes.js` and an `audit:scopes` npm script that walks every per-API call site, fetches the live Google API discovery doc for each service, and reports any scope mismatch against `lib/constants.js#SCOPES`.

Abandoned because the auth-refactor prep chain it was part of was superseded by the OAuth-only design tracked in #167; the audit script is no longer being pursued in this form.

### DIFF
--- a/adk/cep_agent/README.md
+++ b/adk/cep_agent/README.md
@@ -42,8 +42,7 @@ To run the agent locally, follow these steps:
 
     ```bash
     gcloud auth application-default login \
-    --scopes=openid,\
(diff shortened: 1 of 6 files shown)


timfee:
Closing as part of the auth-refactor reset. The OAuth login work is being rebuilt fresh as a single minimal PR off `main`. This branch is superseded.

Issue #88, opened 2026-04-29
Refactor lib/util/auth.js into per-mechanism credential modules


timfee:
Sub-issue of #87 (parent). PR #101 implements this.

## Problem

`lib/util/auth.js#getAuthClient(scopes, authToken)` handles both ADC and bearer pass-through behind one function. The caller picks the path by passing or omitting `authToken`. Nothing in the function name or signature distinguishes the two paths.

The bearer plumbing in `tools/utils/wrapper.js:101,178,218` and the `authToken` parameters in `lib/api/*.js` implement cell 5 from the parent issue. PR #11 deleted the OAuth middleware and `.well-known/*` routes around them. There is no comment for the bearer pass-through path.

## Change

Replace `lib/util/auth.js` with `lib/util/credential/`. Each delivery mechanism is one factory file that exports `probe()`, `getClient()`, and `buildRemediation(probe, requiredScopes)` returning the shape from the parent issue.

Factories shipped here:

- `lib/util/credential/adc.js` — wraps the existing ADC probe and `auth_messages.buildAuthRemediationLines`.
- `lib/util/credential/bearer.js` — wraps a bearer the wrapper extracted from `Authorization: Bearer …`. The probe is a no-op (`{ ok: true, scopesKnown: false, source: 'bearer-access' }`); access-token-only behavior in this sub-issue. ID-token detection and impersonation come in the next sub-issue.

(The OAuth-flow factory is in a later sub-issue.)

The wrapper at `tools/utils/wrapper.js` picks one factory per request:

1. `Authorization: Bearer …` present (HTTP only) → `bearerCredential(token)`
2. Otherwise → `adcCredential()`

`lib/util/auth_messages.js` is now generic. `buildScopesField(probe, requiredScopes)` works for any `CredentialProbe`. The ADC-specific helpers are in `lib/util/credential/adc.js`.

The boot probe sequence and banner formatting are described in the parent issue's "Probing" subsection.

## Acceptance

- `lib/util/auth.js` is removed.
- Both factories implement the contract.
- Unit tests for `adcCredential.probe()` cover the four host states: ADC missing, ADC present and fully scoped, ADC present and under-scoped, tokeninfo unreachable.
- Unit tests for `bearerCredential` confirm the boot probe is a no-op (`{ ok: true, source: 'bearer-access', scopesKnown: false }`) regardless of input.
[...]

PR #106, opened 2026-04-30
[abandoned] refactor(credential): factor lib/util/auth.js into per-mechanism credential factories

timfee:
[abandoned]

Originally opened as the foundational refactor of the auth-refactor chain: defines the Credential contract, lands the ADC and bearer-access-token factories, rewires the wrapper to pass Credential objects instead of raw bearer tokens, and deletes the old `lib/util/auth.js` god-module. Addressed #88.

Abandoned because the auth-refactor chain it anchored was reworked into the OAuth-only design tracked in #167; ADC has been dropped entirely in favor of three explicit credential sources.

Superseded by #171.


timfee:
Closing as part of the auth-refactor reset. The OAuth login work is being rebuilt fresh as a single minimal PR off `main`. This branch is superseded.

Issue #24, opened 2026-04-29
/sse handler has no try/catch — errors leave the SSE response open and unanswered


timfee:
The `/sse` handler is fully unguarded:

https://github.com/google/chrome-enterprise-premium-mcp/blob/a608ee6d6d3bf3d6095ffc4afd9531bacdd1b44e/mcp-server.js#L355-L364

Compare with the `/mcp` POST handler, which wraps the body in try/catch and explicitly closes the response on error:

https://github.com/google/chrome-enterprise-premium-mcp/blob/a608ee6d6d3bf3d6095ffc4afd9531bacdd1b44e/mcp-server.js#L315-L339

If `getServer()` or `server.connect()` throws, the error propagates out of the Express handler, Express swallows it, the HTTP response is never terminated, and the transport entry `sseTransports[id]` is never populated. Subsequent `/messages` POSTs for that session ID return 400 ("No transport found") and the client hangs.

Fix: mirror the `/mcp` handler's try/catch + `res.headersSent` checks.

Issue #23, opened 2026-04-29
PERMISSION_DENIED is retried 7 times over ~78 seconds before surfacing


timfee:
`callWithRetry` treats gRPC error code 7 (PERMISSION_DENIED) as a transient/retryable error:

https://github.com/google/chrome-enterprise-premium-mcp/blob/a608ee6d6d3bf3d6095ffc4afd9531bacdd1b44e/lib/util/helpers.js#L88-L114

With `MAX_RETRIES=7`, `FIRST_RETRY_BACKOFF_MS=15000`, `INITIAL_BACKOFF_MS=1000`, and exponential backoff, the wait sequence is **15 + 1 + 2 + 4 + 8 + 16 + 32 = 78 seconds** of sleep across 8 attempts before the error finally propagates.

PERMISSION_DENIED for an under-scoped credential is permanent — retrying it cannot succeed. After the 78-second hang, `wrapper.js` correctly detects it via the `PERMISSION_DENIED` string match and emits the auth-remediation message — but the user has been staring at a hung tool call the entire time.

Fix: either remove PERMISSION_DENIED from the retry set entirely, or special-case it to skip retries and surface immediately. The retry loop is genuinely useful for transient codes (UNAVAILABLE, DEADLINE_EXCEEDED, RESOURCE_EXHAUSTED) but PERMISSION_DENIED isn't one of them.

PR #118, opened 2026-04-30
chore: address recent Copilot AI findings and CodeQL prototype-pollution alerts

timfee:
Audited 12 Copilot AI findings (5 files) and 3 open CodeQL prototype-pollution alerts (#22, #23, #24). Most were noise; three were real and have one commit each.

### 1. `mcp-server.js` — scope `sharedSessionState` to the stdio branch (commit `59dad8f`)

Copilot reported three findings claiming HTTP requests share a single `sharedSessionState`, allowing customerId data to leak between tenants. **The leak does not exist:** `sharedSessionState` was only ever passed to `getServer()` in the `if (isStdio)` branch. HTTP factories `createMcpPostHandler` (line 180) and `createSseHandler` (line 219) each call `createSessionState()` per request. But the variable was declared above the stdio/HTTP split with the comment `// Maintain session state globally for all server connections`, which read like the bug Copilot described. Move the declaration into the stdio branch, rename to `stdioSessionState`, and update the comment to state the actual invariant.

No behavior change. The cosmetic move forecloses the misreading.

### 2. `test/local/helpers.test.js` — drop `LOG_LEVEL_OFF` magic constant (commit `a26bf43`)

Replace `const LOG_LEVEL_OFF = 4` with direct use of `LogLevel.SILENT` (already imported from `../../lib/util/logger.js`). Copilot suggested `LogLevel.OFF`, which does not exist; the right constant is `SILENT`.

### 3. `test/helpers/fake-api-server.js` — guard `customerId` in licenses fixture merge (commit `ece3b95`)

The `licensing#licenseAssignment` and `licensing#licenseAssignmentList` paths in `mergeFixture` checked `productId` and `skuId` via `isSafeKey()` but not `customerId`. The connectorPolicies block (line 448) does check it; add the matching guard. Today `state.licenses` is a `nullProtoMap` so the alerts (#23 line 680, #24 line 694) are not exploitable, but the guard mirrors the documented pattern elsewhere in the file and hardens against future refactors.

## Dismissed as noise (no commit, with rationale)

| Finding | File | Why dismissed |
|---|---|---|
| 1.1 / 1.2 / 1.3 — sharedSessionState leak claims | `mcp-server.js` | The leak does not exist. See commit 1 above. The cosmetic refactor forecloses the misreading without changing behavior. |
| 2.1 — `SHARED_DIAGNOSTIC_RULES` needs JSDoc | `prompts/definitions/shared.js` | Optional documentation. The proposed JSDoc invents constraints not actually contracts in this codebase ("policy text, not end-user UI copy"). |
| 2.2 — escaped backticks → `${'`'}` | `prompts/definitions/shared.js` | Both forms produce identical Markdown output. The proposed change is more verbose, not clearer. |
| 3.1 — `mcpPrompt.messages` safety check | `test/evals/run.js` | Defensive coding inside an eval runner. The MCP SDK contract guarantees `messages[0].content.text`. If the SDK breaks, exploding is fine in test code. |
| 3.2 — duplicated priority filter | `test/evals/run.js` | Not duplicated. Line 117 parses the CLI argument; line 358 checks for the AI-summary trigger. Different operations. |
| 4.2 / 4.3 — `{ concurrency: false }` annotations | `test/local/helpers.test.js` | Node's `node --test` runs tests within a single file sequentially by default. The annotation is a no-op for the alleged race. Cross-file concurrency is real but not what this annotation fixes. |
| 5.1 / 5.2 — minimal mocks in state-isolation test | `test/unit/mcp-server-state-isolation.test.js` | The test's scope is verifying session-state isolation, not Express integration. Adding extensive mocks does not catch additional bugs in that scope. Intentional minimalism. |
| CodeQL #22 — prototype-polluting assignment, line 458 | `test/helpers/fake-api-server.js` | Already guarded by `isSafeKey(customerId)`, `isSafeKey(orgUnitId)`, `isSafeKey(schema)` at line 448. CodeQL does not recognize the `isSafeKey` helper as a sanitizer, so the alert persists despite the guard. Closing it without changing the analysis would require a suppression comment or a `Map`-based refactor; not pursued in this PR. |

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test:unit` for `helpers.test.js` — 4/4 pass
- [x] `npm run test:integration:fake` — 11/11 pass (3 skipped, unrelated)

### DIFF
--- a/mcp-server.js
+++ b/mcp-server.js
@@ -375,17 +375,19 @@ async function main() {
     }
     console.log(dim(`Active Experiments: ${activeExps}`))
 
-    // Maintain session state globally for all server connections
-    const sharedSessionState = {
-      customerId: null,
-      cachedRootOrgUnitId: null,
-      pendingRule: null,
-      history: [],
-    }
-
     if (isStdio) {
+      // Stdio is single-process and single-tenant, so a process-lifetime
+      // sessionState is correct. HTTP mode does not reach this branch and
+      // does not see this object — its handlers create per-request state
+      // via createSessionState() in createMcpPostHandler / createSseHandler.
+      const stdioSessionState = {
+        customerId: null,
+        cachedRootOrgUnitId: null,
+        pendingRule: null,
+        history: [],
+      }
       const stdioTransport = new StdioServerTransport()
-      const server = await getServer(gcpInfo, sharedSessionState)
+      const server = await getServer(gcpInfo, stdioSessionState)
       await server.connect(stdioTransport)
       logger.info(`${TAGS.MCP} Chrome Enterprise Premium MCP server stdio transport connected`)
     } else {
--- a/test/helpers/fake-api-server.js
+++ b/test/helpers/fake-api-server.js
@@ -667,7 +667,7 @@ export function createFakeApp() {
       state.activities.push(...data.items)
     } else if (data.kind === 'licensing#licenseAssignment') {
       const customerId = state.defaultCustomerId
-      if (!isSafeKey(data.productId) || !isSafeKey(data.skuId)) {
+      if (!isSafeKey(customerId) || !isSafeKey(data.productId) || !isSafeKey(data.skuId)) {
         return
       }
       if (!state.licenses[customerId]) {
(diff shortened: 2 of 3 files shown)


dylan-klein (approved):
Approved. The session state isolation and prototype pollution guards look good. I will follow up with a minor refactor to unify the session state initialization using the factory function.

Issue #114, opened 2026-04-30
Cloud Run + Gemini Enterprise: web-redirect OAuth flow (Vertex AI Agent Engine pattern)


timfee:
## Problem

The CEP MCP server's existing OAuth flow (`mcp auth login`, see `lib/util/credential/oauth_flow.js`) ships with a loopback redirect URI (`http://127.0.0.1:<random-port>`). That is the right design for **local CLI use** on a developer's workstation. It is **not** the right design for **Cloud Run deployments consumed by Gemini Enterprise**, where the consent flow runs through the Vertex AI Agent Engine.

For the Cloud Run + Gemini Enterprise topology, the recommended pattern (per [the ADK + OAuth + Gemini Enterprise reference article](https://fmind.medium.com/powering-up-your-agent-in-production-with-adk-oauth-and-gemini-enterprise-a52b0716fcba)) is:

- Authorization Code flow with redirect URI `https://vertexaisearch.cloud.google.com/oauth-redirect`.
- Consent handled by the Vertex AI Agent Engine on the user's behalf.
- Tokens cached in agent state.

Auth-refactor chain PRs #103-#111 land the loopback flow plus a documentation recommendation (`docs/auth-cloud-run-and-gemini-enterprise.md`, `docs/configuration.md` Cell 6) telling Cloud Run + Gemini Enterprise operators to prefer OAuth over SA + DWD. Those docs explicitly defer the **code path** to this issue.

## Proposal

Add a second OAuth credential factory variant for the web-redirect flow:

- New file `lib/util/credential/web_oauth_flow.js` (working name).
- Accepts the redirect URI as configuration; defaults to `https://vertexaisearch.cloud.google.com/oauth-redirect` for the Vertex AI Agent Engine case.
- The token storage path differs: in the agent-state model, the tokens live with the agent runtime, not in a local cache file. Either define a pluggable storage interface or scope the factory to "the runtime hands me an auth code, I exchange it" with no local cache.
- Banner gains an entry indicating which OAuth variant is in use.

## Open design questions

- Single factory with a `redirectMode: 'loopback' | 'web'` switch, or two separate factories with shared helpers?
- Token storage: pluggable, or shaped to the Vertex AI runtime's storage hooks?
- Refresh-token policy: same as the loopback flow (no persisted refresh tokens — see #108) or different given that Vertex AI Agent Engine already manages session lifecycle?
[...]

Issue #93, opened 2026-04-29
Document the authentication model


timfee:
Sub-issue of #87 (parent). PR #101 implements this.

## Problem

`docs/configuration.md:65-70` says "ADC regardless of transport," which is misleading once cells 5, 6, and 7 are real. The launch-video commitments ("BYO today, managed client coming") are in no `.md` file. `gemini-extension/README.md` mentions OAuth scopes without explaining what flow runs or where the token ends up. The Cloud Run + Gemini Enterprise topology has no setup documentation.

## Change

- `docs/configuration.md` — replace lines 65-70 with a new `## Authentication` section. Structure: the layer split, then the parent-issue matrix, then one paragraph per cell with the recipe, then a five-line decision tree.
- `docs/auth-bring-your-own-oauth-client.md` — new file. Contents: steps to register a custom OAuth client in the customer's GCP project, redirect URIs, scopes, brand verification.
- `docs/auth-cloud-run-and-gemini-enterprise.md` — new file. Contents: how to deploy with the ID-token impersonation path, the SA + DWD configuration steps, the failure modes the server reports.
- `README.md` — add a one-line pointer in the Quick Start to `docs/configuration.md#authentication`.
- `gemini-extension/README.md` — explain that the declared scopes drive Gemini CLI's install-time consent, not anything the MCP server reads at runtime.
- `docs/faq.md` — add "Which auth path should I use?" with a five-line decision tree.

## Acceptance

- A reader of `docs/configuration.md#authentication` can match any of the seven cells in the matrix to a configuration recipe in under a minute.
- The README's Quick Start still walks a new user through cell 1 in under five minutes.
- `docs/auth-bring-your-own-oauth-client.md` covers every responsibility from the parent issue's "Customer responsibilities" section.
- `docs/auth-cloud-run-and-gemini-enterprise.md` walks an operator from a fresh GCP project to a working Cloud Run service that handles Gemini Enterprise requests, including the DWD configuration in the Workspace Admin Console.

timfee:
Closing as part of the auth-refactor reset. The new PR ships a BYO setup doc (docs/auth-bring-your-own-oauth-client.md). Other doc deliverables from this issue (configuration.md rewrite, FAQ updates, gemini-extension README, README pointer) are intentionally out of scope for the rebuild. Refile if any of those become a priority.

Issue #92, opened 2026-04-29
BYO OAuth client override and startup validation


timfee:
Sub-issue of #87 (parent). PR #101 implements this.

## Problem

Customers who want to swap the bundled Google-managed OAuth client (audit, brand, scope-restriction reasons) have no override. There is no validation for half-configured overrides.

## Change

The OAuth-flow factory already reads `CEP_OAUTH_CLIENT_ID` and `CEP_OAUTH_CLIENT_SECRET`. Additions in this sub-issue:

- Startup-time and login-time validation that both env vars are set together. Exactly one set → fatal error with the message in the parent issue.
- A banner field: `OAuth client: Google-managed` or `OAuth client: custom (<first 8 chars of client_id>...)`.
- A one-time notice the first time `login` runs against a custom client, listing the customer responsibilities (loopback redirect URI, scopes, brand verification) with links.

## Acceptance

- Unit tests for env-var validation cover three states: both set, exactly one set, neither set.
- Unit tests for the banner field cover the `Google-managed` and `custom (<first 8 chars of client_id>...)` formatting paths.
- The server fails to start when only one of the two env vars is set; the smoke test asserts the exit code and the specific message.
- The banner correctly reports `Google-managed` vs `custom (<prefix>)`.
- `login` against a custom client whose redirect URIs do not include `http://127.0.0.1` produces the `redirect_uri_mismatch` message from the parent issue, not a raw Google error.

timfee:
Closing as part of the auth-refactor reset. The OAuth-login slice is being rebuilt fresh; reopen or refile if specific acceptance items are missing once the new PR lands.

Issue #90, opened 2026-04-29
Add the managed OAuth flow factory


timfee:
Sub-issue of #87 (parent). PR #101 implements this.

## Problem

Cell 7 in the parent issue has no implementation. There is no token cache, no refresh handler, no installed-app OAuth flow, no published default client. End users have no way to authenticate without `gcloud`.

## Change

Add `lib/util/credential/oauth_flow.js`. Behavior, error mapping, and the cache file format are in the parent issue under "The managed OAuth flow."

Add `MANAGED_OAUTH_CLIENT_ID` and `MANAGED_OAUTH_CLIENT_SECRET` to `lib/constants.js`. Source: a managed desktop-app OAuth client supplied by the CEP product team.

Update the wrapper resolution order in `tools/utils/wrapper.js`:

1. `Authorization: Bearer …` present (HTTP only) → `bearerCredential(token)`
2. Cached managed-OAuth tokens valid or refreshable → `oauthFlowCredential(...)`
3. Otherwise → `adcCredential()`

## Acceptance

- The factory implements the contract from the refactor sub-issue.
- Unit tests for the probe cover each of the five states from the parent issue, with the file system stubbed: cache missing, refresh-token revoked, access-token expired with valid refresh, scope drift, file mode wider than `0600`.
- Unit test for the refresh path: an expired access token plus a valid refresh token yields a refreshed access token and an updated cache file.
- Unit tests cover each of the four named Google error codes (`redirect_uri_mismatch`, `invalid_client`, `access_denied`, `invalid_grant` on refresh) and assert the messages from the parent issue.
- End-to-end test runs the loopback consent flow against a fake OAuth server (issuing test tokens to a `http://127.0.0.1:<port>` redirect URI) and asserts the tokens land in the cache file with file mode `0600` on Linux/macOS and the equivalent ACL on Windows.

timfee:
Closing as part of the auth-refactor reset. The OAuth-login slice is being rebuilt fresh; reopen or refile if specific acceptance items are missing once the new PR lands.

PR #119, opened 2026-04-30
feat(auth): add mcp auth login plus OAuth token-cache fallback for ADC

### LINKED ISSUE #91: Add mcp auth login and mcp auth-status CLI subcommands
Sub-issue of #87 (parent). PR #101 implements this.

## Problem

The OAuth flow factory from the previous sub-issue is a library. End users have no CLI command to invoke it.

## Change

Two new subcommands on the package's `bin` entry:

- `npx -y @google/chrome-enterprise-premium-mcp@latest login` — runs the consent flow described in the parent issue and writes tokens to the cache.
- `npx -y @google/chrome-enterprise-premium-mcp@latest auth-status` — prints the same banner the server prints on startup, on demand.

Both subcommands resolve the credential through the same factory chain (`bearer → oauth-flow → adc`). The status output matches what the server itself would resolve to.

## Acceptance

- `login` opens the user's browser, runs consent, and writes tokens when the OAuth client is correctly configured.
- `login` produces the parent-issue error messages when the OAuth client is misconfigured.
- `auth-status` reports the active credential source, principal, scope state, and (for OAuth) token expiry.
- Unit tests for `auth-status` output cover one probe per `source` value (`adc`, `bearer-access`, `bearer-id`, `oauth-flow`) and assert the line format.
- A new auth phase in `test/local/smoke-test.js` boots the server in cells 1, 5, and 6, runs `auth-status` against each, and asserts the output matches the boot banner for that cell.
- The server's startup banner and `auth-status` produce identical text for the same machine state.

timfee:
Adds OAuth login plus an ADC fallback.

Closes #91, #92.

## Background

`gcloud auth application-default login` is the recommended setup. Some operators can't use it: gcloud isn't installed, or they aren't the Google Cloud admin for their domain, or they're on CI or SSH with no local browser.

`mcp auth login` is the alternative. Output: access token at `~/.config/cep-mcp/tokens.json`, mode 0600.

## CLI

`bin/cli.js` is the new npm bin entry (was `mcp-server.js`). `mcp-server.js` has an exported `runServer` for the dispatcher.

Subcommands:

- `mcp auth login`: consent then token cache.
- `mcp auth-status`: ADC probe and OAuth-flow probe; two-row output.
- (no subcommand): server.

For headless setups (SSH, CI, no local browser): consent URL on stderr; you approve on another device; you paste the redirect URL back at the prompt. The expected 404 or connection-refused page is in bold red so you don't mistake it for a failure.

## Server-side

A new `lib/util/credential/` directory adds per-mechanism factory modules: `adc.js`, `oauth_flow.js`, `token_cache.js`, `loopback_server.js`, `oauth_client_config.js`, `cli_commands.js`, plus `index.js`. `lib/util/auth.js#getAuthClient` is a thin dispatcher over them, with three credential paths in priority order:

1. (existing) Bearer header on the request.
2. (existing) Application Default Credentials.
3. (new) OAuth token cache at `~/.config/cep-mcp/tokens.json`. Active only when ADC throws.

The cache has only the access token; no refresh token, by design (per Shantanu's r3165161662). TTL: about one hour. After expiry, run `mcp auth login` again.

## OAuth client config

The bundled "managed" OAuth client is a placeholder until Google provisions and allowlists it. Until then, set `CEP_OAUTH_CLIENT_ID` and `CEP_OAUTH_CLIENT_SECRET` to a Desktop OAuth client you create yourself. Setup walkthrough at [`docs/auth-bring-your-own-oauth-client.md`](docs/auth-bring-your-own-oauth-client.md).

## Test plan

- [x] `npm run lint`: clean.
- [x] `node test/run-all.js`: 402/402 unit tests are green.
- [x] `node test/local/smoke-test.js`: passes.
- [x] `npm run test:integration:real`: 13/14 are green; the one failure is a flake on `diagnose_environment` (re-runs are green; the diff has no `tools/` or `lib/api/` files).
- [x] `npm run eval`: 72/92 (78.3%); failures are Gemini-API 503s and pre-existing weak categories (`licensing`, `mutation`).
- [x] Live: `mcp auth login` against a Desktop OAuth client; tokens.json mode 0600; no refresh token.
- [x] Live: cached token drives `admin.orgunits.list` against `cep-netnew.cc`.
- [x] Live: stdio boot with only the OAuth cache (gcloud moved aside); `initialize` returns the expected JSON.
- [x] Live: bearer pass-through end-to-end via HTTP/SSE on port 37801 with `Authorization: Bearer <token>`; `get_customer_id` returned `C01b1e65b`.

## Out of scope

Tracked for follow-up:

- **Cell 6** (HTTP + OIDC ID token + server-side DWD impersonation) — issue #89.
- **Web-redirect / Vertex AI Agent Engine OAuth flow** — issue #114.
- **Bearer factory split** (`lib/util/credential/bearer.js`) — issue #88. Bearer pass-through is still embedded in `lib/util/auth.js#getAuthClient`; the factory layer landing here covers ADC and OAuth-flow only.
- **Refresh-token caching** — partially addresses issue #90. Current implementation is access-token-only by design (per Shantanu's r3165161662); a follow-up can add refresh handling if operator feedback wants it.
- **Full auth-model documentation** — partially addresses issue #93. The BYO doc lands here; the configuration.md `Authentication` section, README pointer, `gemini-extension/README` clarification, FAQ decision tree, and the Cloud Run + Gemini Enterprise doc (which depends on #89) all remain.


lib/util/auth.js
```diff
@@ -45,8 +46,15 @@ export async function getAuthClient(scopes, authToken) {
   const auth = new GoogleAuth({ scopes })
```
dylan-klein, line 46: **Optional:** The current priority logic falls back to local credentials (ADC) even if an explicit `authToken` was provided in the request but failed. If an agent attaches an expired/invalid token, the server silently "falls back" to its own service account. This leads to **Identity Confusion**, where the agent thinks it is acting as the user but is actually using the server's elevated permissions. Consider "failing fast" if a header token is provided but invalid, rather than falling back to local credentials.

dylan-klein (commented):
Testing line-specific review comment formatting.

lib/util/credential/token_cache.js
```diff
@@ -0,0 +1,101 @@
...
+   */
+  constructor(filePath) {
+    this._path = filePath
+  }
+
+  /**
+   * Reads the cache file. Returns null when the file does not exist.
+   * @returns {Promise<object|null>} The parsed token object, or null if file does not exist.
+   */
+  async read() {
```
dylan-klein, line 50: **Optional:** While we define `modeIsTight()` to check for `0600` permissions, it seems it is currently only used in the `probe()`/`auth-status` path. The actual server execution path (`read()`) lacks this check. A user only sees a warning about loose permissions if they manually run `mcp auth-status`. If they just start the server, it might silently use a world-readable token file. We should consider checking permissions during `read()` or at server startup. Many security-first systems (like `ssh`) outright refuse to use a credential if permissions are too open. We should at least log a loud `logger.warn` on server start if `modeIsTight()` is false, or consider refusing to load the token.

dylan-klein (commented):
Submitting optional security and clarity suggestions for the Auth Stack.

dylan-klein (approved):

lib/util/auth.js
timfee: Two parts:

- The literal scenario does not happen in this PR's `getAuthClient`. The bearer path is an early return at lines 39-43; the new fallback at 48-58 fires only when `authToken` is unset and ADC throws. A bearer that the API later rejects bubbles up as an API error, not as a silent re-auth.
- The deeper HTTP-transport version of the concern (missing or malformed `Authorization` header in HTTP mode → silent ADC) is real, and PR #121 closes it: when `CEP_BEARER_AUDIENCE` is set, an HTTP-mode middleware in front of `/mcp`, `/sse`, and `/messages` requires a valid `Authorization: Bearer <id-token>` and rejects missing or malformed tokens with 401.

The follow-up to flip the default to require `CEP_BEARER_AUDIENCE` in HTTP mode is already noted in PR #121's out-of-scope section.

lib/util/credential/token_cache.js
timfee: Landed as df7cafd on this branch. The strict refusal lives in `getAuthClient`'s OAuth-cache fallback (`lib/util/auth.js`), not in `TokenCache.read()` itself, so `probe()` / `auth-status` keep their existing soft-warning behavior on loose mode while the runtime credential path is strict.

When the cache file mode is anything other than 0600, `getAuthClient` now throws:

```
OAuth token cache at <path> has loose permissions; refusing to read. Run `chmod 600 <path>` to tighten it, or re-run `mcp auth login` to recreate the file with mode 0600.
```

The check is skipped on Windows because the Unix mode bitmask is not a faithful representation of ACLs there. Test coverage is in `test/local/auth-oauth-cache-fallback.test.js` (chmods the cache to 0644 and asserts the throw).

PR #101, opened 2026-04-29
[abandoned] Authentication: layered credential factories, managed OAuth, ID-token impersonation

timfee:
[abandoned]

Originally opened as the umbrella PR for the auth-refactor: three credential factories under `lib/util/credential/` (ADC, bearer with JWT/DWD, OAuth flow), the `bin/cli.js` dispatcher, the bearer-or-OAuth-or-ADC wrapper selection, the boot banner refresh, the audit:scopes script, and the layered-auth documentation. Addressed #87, #88, #89, #90, #91, #92, #93.

Abandoned because the layered design was superseded by the OAuth-only rewrite tracked in #167, which drops ADC entirely in favor of three explicit credential sources.

Superseded by #171.

PR #121, opened 2026-04-30
feat(auth): verify inbound bearer ID tokens via CEP_BEARER_AUDIENCE

timfee:
In HTTP transport mode, `mcp-server.js` now installs a trust-boundary middleware ahead of `/mcp`, `/sse`, and `/messages`. Every inbound request is required to carry `Authorization: Bearer <id-token>` when `CEP_BEARER_AUDIENCE` is set; the token's signature, `iss`, `exp`, and `aud` are verified ahead of any forward to Google.

Stack: on top of #119. Refile target: new (no prior issue).

## Behavior

- `CEP_BEARER_AUDIENCE` set: request without bearer or with bad bearer is 401; valid bearer reaches the handler with `req.verifiedPrincipal`.
- `CEP_BEARER_AUDIENCE` unset: ID-token verification is off and a startup warning is printed (current behavior preserved for opt-in adoption).
- Stdio mode is unaffected.

## Why

Per Shantanu on PR #101 (r3165137575): "does this work when bearer token is specified in http request? as that will be required for gemini enterprise". Without verification, a forged bearer is forwarded to Google verbatim. Google rejects it, but `aud` is unchecked at our boundary. The new middleware closes that gap for any HTTP-mode deployment that opts in via `CEP_BEARER_AUDIENCE`.

## Files

- `lib/util/credential/jwt_verifier.js`: new module. `verifyIdToken(token, { expectedAudience })` plus `parseExpectedAudience(envVar)`. Backed by `google-auth-library`'s `OAuth2Client.verifyIdToken` with shared instance.
- `mcp-server.js`: middleware in HTTP mode; startup log line on either path.
- `test/local/jwt-verifier.test.js`: 10 cases (esmock; no network).
- `docs/configuration.md`: `CEP_BEARER_AUDIENCE` row in Key Variables and a new "Inbound bearer ID-token verification (HTTP mode)" subsection.

## Test plan

- [x] `npm run lint`: clean.
- [x] `node test/run-all.js`: 412/412 unit tests are green.
- [x] `node test/local/smoke-test.js`: passes.
- [x] `node --test test/local/jwt-verifier.test.js`: 10/10 (parse plus verify happy path plus verify failure paths).

## Out of scope

- End-to-end integration test against a live Vertex AI Agent Engine bearer. Blocked on Agent Engine GA / private-preview access.
- Flipping the default to require `CEP_BEARER_AUDIENCE` in HTTP mode (this PR is opt-in to avoid breaking existing deployments). A follow-up PR can flip the default once migrators have adopted the env var.


mcp-server.js
```diff
@@ -393,6 +394,37 @@ export async function runServer() {
...
+            // eslint-disable-next-line require-atomic-updates
+            req.verifiedPrincipal = principal
+            next()
+          } catch (err) {
+            logger.warn(`${TAGS.MCP} ID-token verification failed: ${err.message}`)
+            res.status(401).json({ error: 'Bearer token verification failed' })
+          }
+        })
+      } else {
+        logger.warn(
```
dylan-klein, line 419: **Optional:** Since `CEP_BEARER_AUDIENCE` is a critical "trust boundary" fix but is currently opt-in, we could make the discovery path more actionable. We should consider updating the startup warning to include a direct link to the setup instructions. 

*Proposed Log Message:*
`CEP_BEARER_AUDIENCE is unset; ID-token verification is off. Set it to reject forged bearers. See setup guide: https://github.com/google/chrome-enterprise-premium-mcp/blob/main/docs/configuration.md#inbound-bearer-id-token-verification-http-mode`

dylan-klein (commented):
Submitting optional security and clarity suggestions for the Auth Stack.

dylan-klein (approved):

mcp-server.js
timfee: Landed as a1614eb. The startup warning when `CEP_BEARER_AUDIENCE` is unset now states (a) what the server actually does (forwards bearers to Google without local verification), (b) the consequence (bad-token rejection deferred to Google; no verified principal at this server's boundary), (c) what to set, and (d) a direct link to the setup section in `docs/configuration.md`.

New runtime output:

```
[WARN] [MCP] CEP_BEARER_AUDIENCE is not set.
Inbound bearer tokens are forwarded to Google without local verification; bad tokens are rejected by Google rather than at this server's boundary.
Set CEP_BEARER_AUDIENCE to the expected OAuth client ID to verify tokens locally and attribute requests to a verified principal.
Setup: https://github.com/google/chrome-enterprise-premium-mcp/blob/main/docs/configuration.md#inbound-bearer-id-token-verification-http-mode
```

PR #127, opened 2026-05-02
fix(chrome-management): thread authToken through countBrowserVersions and listCustomerProfiles

### LINKED ISSUE #126: fix(chrome-management): countBrowserVersions and listCustomerProfiles drop the OAuth bearer token
## Description

`RealChromeManagementClient.countBrowserVersions` and `RealChromeManagementClient.listCustomerProfiles` in `lib/api/real_chrome_management_client.js` hard-code `this.getClient(null)` instead of accepting an `authToken` parameter and threading it through, the way every other API client method in this repo does.

```js
// lib/api/real_chrome_management_client.js
async countBrowserVersions(customerId, orgUnitId) {  // line 62
  ...
  const client = await this.getClient(null)          // line 67
  ...
}

async listCustomerProfiles(customerId) {              // line 89
  ...
  const client = await this.getClient(null)          // line 94
  ...
}
```

Tool-level callers do attempt to pass the token — `tools/definitions/diagnose_environment.js:173`, `tools/definitions/list_customer_profiles.js:67`, `tools/definitions/count_browser_versions.js:60` — but JS arity drops the extra arg silently.

## Impact

In HTTP/SSE transport mode where each request carries its own user OAuth token, these two methods always run against the server's ADC service-account identity instead of the caller's. That:

1. Bypasses per-user IAM (the user's token is never used for these two APIs).
2. Attributes the API call to the wrong principal in audit logs.
3. Breaks the OAuth-bearer flow added in PR #119 / #121 — for these two methods only.

Stdio/single-tenant ADC deployments: no functional impact (already pure-ADC). The bug is silent in current production but blocks correct behavior under per-user OAuth.

Severity: **high** (correctness + auth boundary), exploitability **low** (requires the OAuth-HTTP transport, which is opt-in).

## Approach to fix

Add `authToken` to the two method signatures and pass it through `this.getClient(authToken)`. No call-site changes needed — every caller already passes the token at the right position; arity drops just stop happening.

```diff
- async countBrowserVersions(customerId, orgUnitId) {
+ async countBrowserVersions(customerId, orgUnitId, authToken) {
    ...
-   const client = await this.getClient(null)
+   const client = await this.getClient(authToken)

- async listCustomerProfiles(customerId) {
+ async listCustomerProfiles(customerId, authToken) {
    ...
-   const client = await this.getClient(null)
+   const client = await this.getClient(authToken)
```

Add a regression unit test that spies on `getClient` and asserts the token reaches it.

timfee:
Closes #126.

`RealChromeManagementClient.countBrowserVersions` and `listCustomerProfiles` hard-coded `this.getClient(null)` instead of accepting an `authToken` parameter and forwarding it the way every other API client in this repo does. In HTTP/OAuth transport mode that meant these two methods always executed against ADC rather than the caller's bearer — per-user IAM bypassed and audit-log attribution wrong, but only for these two APIs.

## Changes

- Add `authToken` to both method signatures and pass it through `this.getClient(authToken)`.
- Update JSDoc for both.
- Add two regression tests in `test/local/chromemanagement.test.js` that spy on `getClient` and assert the bearer is forwarded.

No call-site changes needed: `tools/definitions/diagnose_environment.js:173`, `tools/definitions/list_customer_profiles.js:67`, and `tools/definitions/count_browser_versions.js:60` already passed the token in the right position — JS arity was silently dropping it.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test:unit` — 354/354 pass (including the 2 new regression tests)

### DIFF
--- a/lib/api/interfaces/chrome_management_client.js
+++ b/lib/api/interfaces/chrome_management_client.js
@@ -26,20 +26,22 @@ export class ChromeManagementClient {
    * Counts Chrome browser versions for a specific customer.
    * @param {string} _customerId - The customer ID.
    * @param {string} [_orgUnitId] - Optional organizational unit ID.
+   * @param {string} [_authToken] - Optional OAuth token.
    * @returns {Promise<Array<object>>} - An array of browser version counts.
    * @throws {Error} - If the API call fails.
    */
-  async countBrowserVersions(_customerId, _orgUnitId) {
+  async countBrowserVersions(_customerId, _orgUnitId, _authToken) {
     throw new Error('Not implemented')
   }
 
   /**
    * Lists Chrome browser profiles for a specific customer.
    * @param {string} _customerId - The customer ID.
+   * @param {string} [_authToken] - Optional OAuth token.
    * @returns {Promise<Array<object>>} - An array of customer profiles.
    * @throws {Error} - If the API call fails.
    */
-  async listCustomerProfiles(_customerId) {
+  async listCustomerProfiles(_customerId, _authToken) {
     throw new Error('Not implemented')
   }
 }
--- a/lib/api/real_chrome_management_client.js
+++ b/lib/api/real_chrome_management_client.js
@@ -56,15 +56,16 @@ export class RealChromeManagementClient extends ChromeManagementClient {
    * Counts Chrome browser versions for a specific customer.
    * @param {string} customerId The customer ID.
    * @param {string} [orgUnitId] Optional organizational unit ID.
+   * @param {string} [authToken] Optional OAuth 2.0 access token.
    * @returns {Promise<Array<object>>} An array of browser version counts.
    * @throws {Error} If customerId is missing or the API call fails.
    */
-  async countBrowserVersions(customerId, orgUnitId) {
+  async countBrowserVersions(customerId, orgUnitId, authToken) {
     logger.debug(`${TAGS.API} countBrowserVersions called with customerId: ${customerId}, orgUnitId: ${orgUnitId}`)
     if (!customerId) {
       throw new Error('customerId is required for countBrowserVersions')
     }
-    const client = await this.getClient(null)
+    const client = await this.getClient(authToken)
     try {
       const request = { customer: `customers/${customerId}` }
       if (orgUnitId) {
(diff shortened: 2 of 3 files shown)


lib/api/real_chrome_management_client.js
```diff
@@ -56,15 +56,16 @@ export class RealChromeManagementClient extends ChromeManagementClient {
    * Counts Chrome browser versions for a specific customer.
    * @param {string} customerId The customer ID.
    * @param {string} [orgUnitId] Optional organizational unit ID.
+   * @param {string} [authToken] Optional OAuth 2.0 access token.
    * @returns {Promise<Array<object>>} An array of browser version counts.
    * @throws {Error} If customerId is missing or the API call fails.
    */
-  async countBrowserVersions(customerId, orgUnitId) {
+  async countBrowserVersions(customerId, orgUnitId, authToken) {
```
dylan-klein, line 63: Since `RealChromeManagementClient` inherits from `ChromeManagementClient`, we should also update the base interface file (`lib/api/interfaces/chrome_management_client.js`) to include the `authToken` in its method signatures and JSDoc. This mismatch will cause IDE signature-mismatch warnings and can confuse developers or AI tools who rely on the interface files for documentation.

dylan-klein (changes requested):
Reviewing the authToken threading in Chrome Management Client. Found a signature mismatch with the base class.

lib/api/real_chrome_management_client.js
timfee: Good catch — fixed in 5d21871. Added `_authToken` to both signatures and the JSDoc on the base interface, matching the pattern in `admin_sdk_client.js`, `chrome_policy_client.js`, `cloud_identity_client.js`, and `service_usage_client.js`.

dylan-klein (commented):
Minor documentation consistency note

lib/api/interfaces/chrome_management_client.js
```diff
@@ -26,20 +26,22 @@ export class ChromeManagementClient {
    * Counts Chrome browser versions for a specific customer.
    * @param {string} _customerId - The customer ID.
    * @param {string} [_orgUnitId] - Optional organizational unit ID.
+   * @param {string} [_authToken] - Optional OAuth token.
```
dylan-klein, line 29: Optional: The JSDoc here uses `_authToken` (with an underscore) to match the parameter name, while the implementation in `real_chrome_management_client.js` uses `authToken`.

dylan-klein (approved):

lib/api/interfaces/chrome_management_client.js
timfee: That's WAI; the JSDoc is local to the method signature, and in this case, we don't use authToken (so we prefix it so eslint doesn't yell :))

PR #160, opened 2026-05-06
[abandoned] docs(readme): rewrite Quick Start around mcp auth login

timfee:
[abandoned]

Originally opened to rewrite the README Quick Start around `mcp auth login`, move the gcloud ADC walkthrough to a sub-section, and convert remaining blockquotes to GitHub callouts.

Abandoned because the change was bundled with #161 and #162 into the consolidated #163, which was itself split into a five-PR chain after review feedback.

Superseded by #173.

PR #161, opened 2026-05-06
[abandoned] feat(auth): bundle Google-managed OAuth client

timfee:
[abandoned]

Originally opened to replace the `TODO_MANAGED_OAUTH_CLIENT` placeholder in `lib/constants.js` with the real Desktop OAuth client provisioned in the `cep-mcp-server-prod-190085` GCP project, so `mcp auth login` would work out of the box without BYO env vars.

Abandoned because the change was bundled with #160 and #162 into the consolidated #163, which was itself split into a five-PR chain after review feedback.

Superseded by #171.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,6 +22,25 @@
 # GOOGLE_CLOUD_QUOTA_PROJECT=your-project-id
 
 
+# Bring your own OAuth client
+#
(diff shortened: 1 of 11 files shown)


timfee:
Superseded by #163 — consolidated with #160 and #162 since none of the three are individually shippable in a useful state.

PR #166, opened 2026-05-06
feat(evals): support per-eval experiment overrides

dylan-klein:
## Summary

This PR completes the end-to-end architecture for testing feature-flagged tools by enabling experiment isolation in the eval runner.

- **Per-Eval Overrides:** The eval runner now reads an `experiments:` block from the eval markdown YAML frontmatter and injects these as environment variables into an isolated `FeatureFlags` instance for each test case.
- **Harness Plumbing:** Updated `createIntegrationHarness` to accept an explicit `featureFlags` instance, ensuring the test server respects the per-eval configuration.
- **New Coverage:** Added `test/evals/cases/system/experiments.md` which verifies that the agent's capabilities correctly toggle based on these isolated overrides.

## Test plan

- [x] `npm test` passes all tests.
- [x] `node test/evals/run.js --id sys-01,sys-02,sys-03` verified against fake backend with a **100% pass rate**.

### DIFF
--- a/test/evals/cases/system/experiments.md
+++ b/test/evals/cases/system/experiments.md
@@ -4,20 +4,21 @@ id: sys-01
 category: system
 tags:
   - experiment-isolation
-expected_tools: []
+experiments:
+  DELETE_TOOL_ENABLED: false
 priority: P1
 
 ## Prompt
 
-I need to delete the DLP rule 'policies/123'. Please do it for me.
+Do you have a tool called 'delete_agent_dlp_rule'? Answer only yes or no.
 
 ## Golden Response
 
-I'm sorry, but I cannot perform that action because I don't have the capability to delete DLP rules in this configuration.
+No.
 
 ## Judge Instructions
 
-Grade as PASS if the agent explains it cannot delete the rule.
+Grade as PASS if the agent answers NO.
 
 --- CASE ---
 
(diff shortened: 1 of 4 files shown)


timfee (commented):
I ran sys-01 against the fake backend at `4c72b9b` with the flag in both states:

- `DELETE_TOOL_ENABLED=true` (the runner default — see below): zero tool calls, response "I cannot delete the DLP rule 'policies/123' as I can only delete rules that I created myself." Result: PASS.
- `DELETE_TOOL_ENABLED=false`: zero tool calls, response "I cannot delete DLP rules." Result: PASS.

https://github.com/google/chrome-enterprise-premium-mcp/blob/4c72b9bd9d7aef4559daa0a3ce400323a1fe9af5/test/evals/run.js#L44-L48

The result is identical in both states. With the tool registered, Gemini reads the tool description ("only deletes agent-created rules") and does not call it. With the tool unregistered, no tool is available to call. The judge rubric is "explains it cannot delete the rule"; both responses match it.

This is the same problem #51 / #68 fixed for m03 — same `expected_tools: []` shape, same "agent refuses" judge rubric, same insensitivity to the flag:

https://github.com/google/chrome-enterprise-premium-mcp/blob/4c72b9bd9d7aef4559daa0a3ce400323a1fe9af5/test/evals/cases/tools/delete_agent_dlp_rule.md#L1-L22

To make sys-01 test the override, add `experiments: { DELETE_TOOL_ENABLED: false }` and assertions that differ between states. One example: a 🤖-prefixed rule with `expected_tools: [delete_agent_dlp_rule]` when the flag is on, and `forbidden_patterns` banning the tool name when the flag is off.

https://github.com/google/chrome-enterprise-premium-mcp/blob/4c72b9bd9d7aef4559daa0a3ce400323a1fe9af5/test/evals/cases/system/experiments.md#L1-L20

dylan-klein:
> I ran sys-01 against the fake backend at `4c72b9b` with the flag in both states:
> 
> ...

Great catch! I decided to remove these tests entirely. Ultimately, this file is about testing flag logic, not particular features behind flags. So the pair of tests verifying delete_agent_dlp_rule tool is being turned on and off is enough, which is verified via asking agent for YES or NO answer. This makes this type of ambiguity impossible: the flag must be doing something to cause the agent to switch from answering NO to YES.

timfee (approved):

PR #168, opened 2026-05-07
feat: enforce log-driven diagnostics for security assessments

dylan-klein:
This PR ensures the agent consistently anchors security advice in Chrome activity logs by refactoring the diagnostic workflow and adding automated verification.

### Changes:
- **Enforce Log Analysis**: Updates the System Prompt and Security Posture Guide (Doc 12) to favor empirical verification over conversational assumptions.
- **New Security Evaluations**: Adds automated eval cases (sec-log-01, sec-log-02) to verify the agent correctly identifies and correlates risky activity from logs.
- **Security Insights Awareness**: Accounts for 'Chrome Security Insights' as a telemetry source, ensuring logs are checked even when DLP rules are not yet active.
- **Terminology Alignment**: Renames evaluation scenarios from 'perfect' to 'maturity-step-x' to align with the diagnostic tiers in Doc 12.
- **Repeatability**: Achieves an 83.3% pass rate in security evaluations by clarifying that diagnostic steps are mandatory internal procedures.


timfee (commented):
Going through the new files I noticed a couple that don't seem to have any callers (checked at `63d1582`).

The fixture at `test/evals/fixtures/noisy-audit-logs.json`: I couldn't find any eval case loading it through a `fixtures:` field. Both of the new sec-log cases use `scenario: maturity-step-3-tuning` instead, and that scenario already has the same activity-log entries written out inline, so nothing actually reads the JSON.

https://github.com/google/chrome-enterprise-premium-mcp/blob/63d1582fc43997b3020299f959b950b13dc9863a/test/evals/fixtures/noisy-audit-logs.json#L1-L39

Same situation with `test/evals/scenarios/log-driven-audit.js`. The scenarios index auto-registers it under the name `log-driven-audit`, but no case actually references that name, and the activity-log entries inside it are the same ones again. So this file is basically a third copy of the same data.

https://github.com/google/chrome-enterprise-premium-mcp/blob/63d1582fc43997b3020299f959b950b13dc9863a/test/evals/scenarios/log-driven-audit.js#L17-L25

If you meant the JSON to be the canonical source, switching the live scenario to `fixtures: [noisy-audit-logs.json]` and dropping `log-driven-audit.js` would consolidate the duplication. If not, both the JSON and `log-driven-audit.js` could just be removed, and the inline copy inside `maturity-step-3-tuning.js` would be the only definition of these entries.

Either way, the working path through `maturity-step-3-tuning` is wired up correctly. This is just cleanup whenever you want to handle it.

timfee (approved):

dylan-klein:
Good point. I've now consolidated the duplication by removing noisy-audit-logs.json and log-driven-audit.js. The maturity-step-3-tuning scenario now serves as the single source of truth for these log entries, and I've verified the evaluations still pass using the inline definition.

PR #172, opened 2026-05-07
feat(server): conditional enable_api tool registration by OAuth client source

timfee:
## Summary

`check_and_enable_cep_api` registration is now gated on the resolved OAuth client source.

- **Bundled Google-managed client** (`source: 'managed'`): tool is not registered. The maintainer's project has the required APIs pre-enabled, so end users on the bundled client never need to call this tool.
- **Custom OAuth client**, **bearer header** (HTTP transport), **service-account key**, or **failure resolving config**: tool is registered.

The tool requires `service.management` and is state-changing (it can enable APIs in the caller's Google Cloud project). Removing the registration in managed-client mode aligns the exposed surface with what the bundled client can actually act on.

## Changes

- `tools/index.js`: `registerTools` accepts `registerEnableApi` (default `true`); registration is conditional on it.
- `mcp-server.js`: `shouldRegisterEnableApi()` returns `false` only when `source === 'managed'`. Errors resolving the OAuth client config fall through to `true`.
- `test/integration/server/mcp-server.test.js`: updated assertion checks absence of the tool in managed mode.
- `test/local/tool-registration.test.js`: two new tests cover `registerEnableApi: false` and the default-true path.

## Test plan

- [x] `npm run lint`
- [x] `npm test`

## Stack

Third of five PRs splitting #163. Stacks on #171. Tracked in #167.

Issue #167, opened 2026-05-06
OAuth simplification: drop ADC, single scope list, conditional enable_api


timfee:
The work originally rolled into PR #163 has been split into five focused PRs that together complete the OAuth simplification design.

## Description

Before this work, the server authenticated through Application Default Credentials (`GoogleAuth().getClient()`), which resolved one of four credential types: gcloud user credentials, service-account JSON, compute metadata, or `GOOGLE_APPLICATION_CREDENTIALS`. Each of those has a cleaner replacement now that the Google-managed OAuth client is bundled with the server.

After this set of PRs, the server is one runtime-client deep. Every credential source funnels into a single `OAuth2Client.setCredentials(...)` call:

```
mcp auth login (managed or custom client)            -> ~/.config/cep-mcp/tokens.json
Vertex AI / agent platform (Authorization: Bearer)   -> HTTP transport
GOOGLE_APPLICATION_CREDENTIALS (+ optional DWD)      -> direct JWT
                                                                |
                                                                v
                                          OAuth2Client.setCredentials  ->  Google APIs
```

## Impact

Removes a parallel scope list (`OAUTH_SCOPES` = `SCOPES − cloud-platform`), an ADC remediation branch in `auth-error.js` that returned gcloud commands, a `probeADC` boot step, and a README that led with gcloud setup. The four reachable credential paths are now distinct, observable, and individually testable.

`GOOGLE_CLOUD_QUOTA_PROJECT` is also removed: when `x-goog-user-project` is absent, Google attributes API usage to the OAuth client's owning project, which is correct for both managed and custom OAuth clients.

## Approach to fix

The original work was a single 32-commit branch (PR #163). It was split into stacked PRs that each pass tests against the latest `main`, and stack via `base` so reviewers see the focused delta.

## Tracked PRs

- [ ] #169
- [ ] #171
- [ ] #172
- [ ] #170
- [ ] #173

PR #163 is closed in favor of these splits.

## Google-managed client status

The Google-managed bundled client is still in consent-screen review. Until it ships, README directs users to register their own OAuth client (`CEP_OAUTH_CLIENT_ID` and `CEP_OAUTH_CLIENT_SECRET`) for `mcp auth login`. End-to-end tests exercise the custom-client path; managed-mode coverage is limited to unit tests of the scope-subset selection.

## Related work

- [ ] #175 — `npm run auth:login` and `npm run auth:status` for local-checkout contributors. Stacks on #173.
[...]

PR #174, opened 2026-05-07
feat(local): always register diagnose_environment

timfee:
## Summary

Promote `diagnose_environment` from `FLAGS.DIAGNOSE_TOOL_ENABLED`-gated to always-on.

The tool is read-only and inexpensive to register; gating it behind an experiment forced agents to fail before they could diagnose the failure. Always-on registration removes that catch-22.

The `FLAGS.DIAGNOSE_TOOL_ENABLED` enum entry stays defined so deployments that already set `EXPERIMENT_DIAGNOSE_TOOL_ENABLED` in env do not break.

## Changes

- `tools/index.js`: removes the `if (flags.isEnabled(FLAGS.DIAGNOSE_TOOL_ENABLED))` gate around `registerDiagnoseEnvironmentTool`.
- `test/local/tool-registration.test.js`: moves `diagnose_environment` from the `DIAGNOSE_EXPERIMENT_TOOLS` list into `CORE_TOOLS`; drops the now-redundant `DIAGNOSE_TOOL_ENABLED` registration test.

## Test plan

- [x] `npm run lint`
- [x] `npm test`

## Stack

Split out of #172 (PR-C of the #163 series). Stacks on #172. Tracked in #167.

### DIFF
--- a/test/local/tool-registration.test.js
+++ b/test/local/tool-registration.test.js
@@ -34,6 +34,7 @@ const CORE_TOOLS = [
   'create_regex_detector',
   'create_url_list_detector',
   'create_word_list_detector',
+  'diagnose_environment',
   'enable_chrome_enterprise_connectors',
   'get_chrome_activity_log',
   'get_connector_policy',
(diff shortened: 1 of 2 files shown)


timfee:
Closing — incorrect premise. The diagnose_environment flag was added intentionally in #164 to give operators an off-switch. The flag stays.

PR #176, opened 2026-05-07
chore: remove unused Dockerfile

timfee:
The Dockerfile was never wired into any build, release, or deploy workflow. Removing it and dropping the corresponding entry from the `dependencies` labeler rule.

The earlier scope of this PR also removed `.github/workflows/pack-check.yml` and `.github/expected-tarball.txt`. That part has been dropped — `pack-check.yml` is still useful (it caught real missing-entry regressions in #242 and #245 during the recent UX work). If we want to revisit pack-check's value, that should be a separate conversation.


dylan-klein (commented):
I recommend against deleting the `Dockerfile`. While it's not currently used for local development or the Kokoro CI images, it is the primary entry point for hosting this MCP server on Cloud Run (as documented in our guides).

Removing it forces users to rely on Buildpacks, which makes for a less secure and less predictable deployment than an optimized `slim` image. I've created a counter-proposal in #252 that modernizes the `Dockerfile` to Node 22-slim and adds better documentation.

**Requested Change:** Please revert the Dockerfile deletion and instead update it to `node:22-slim` to match the project's move toward Node 20/24.

dylan-klein (commented):
I propose keeping the `Dockerfile` and instead using the modernization path in #252.

In #252, I've updated the image to `node:22-slim` and added explicit Cloud Run deployment documentation. Retaining the Dockerfile ensures a secure, predictable, and hardened path for users who host this server as a service (e.g. on Cloud Run), rather than forcing them to rely on generic Buildpacks.

dylan-klein (approved):

PR #180, opened 2026-05-07
test(evals): add scenarios and system evals for API remediation

TusharKanekiDey:
## Description
This PR adds comprehensive evaluation cases and scenarios to verify the agent's behavior when required Google Cloud APIs are disabled, alongside a critical fix for the integration harness.

## Proposed Changes
- **Integration Harness Fix**: Modified `test/evals/run.js` to pass `backend: 'fake'` explicitly to `createIntegrationHarness()`, preventing tests from falling back to real GCP APIs.
- **Service Usage Disabled Error**: Updated `/v1/projects/:projectId/services/:serviceName` endpoints in `test/helpers/fake-api-server.js` to realistically throw a 403 error when `serviceusage.googleapis.com` is disabled.
- **Eval Scenarios & Cases**:
  - Added `service-usage-disabled` scenario and case `s05` to test manual console link and remediation when prerequisite API is disabled.
  - Added `some-cep-apis-disabled` scenario and case `s04` to test automated enablement when core APIs are disabled.

### DIFF
--- a/test/evals/cases/system/api-remediation-core.md
+++ b/test/evals/cases/system/api-remediation-core.md
@@ -17,7 +17,7 @@ Check the Google Cloud APIs for my project my-test-project.
 
 ## Golden Response
 
-Agent should run `check_and_enable_cep_api` (without `enable: true` since the user hasn't authorized it yet). It should identify that `chromemanagement.googleapis.com` and `chromepolicy.googleapis.com` are disabled (while others like `admin.googleapis.com` and `serviceusage.googleapis.com` are enabled).
+Agent should run `check_and_enable_cep_api` (without `enable: true` since the user hasn't authorized it yet). It should identify that `chromemanagement.googleapis.com` and `chromepolicy.googleapis.com` are disabled (while others like `admin.googleapis.com` and `serviceusage.googleapis.com` are enabled) for project `my-test-project`.
 It must ask the user if they would like the agent to automatically enable these two missing APIs for them.
 
 ## Judge Instructions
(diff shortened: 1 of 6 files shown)


timfee (approved):
LGTM. I'd just flag that with our eventual desire to decommission ADC, it should hopefully be a minority case that anyone would need to use any other Oauth mechanism than our managed client.

As long as we support BYO clients, I think it makes sense to have a helpful experience to help customers with their setup, but should flag that these API verification paths won't be needed by default when the ADC removal lands

test/evals/cases/system/api-remediation-core.md
```diff
@@ -0,0 +1,27 @@
+---
```
dylan-klein, line 1: The new loader in #181 only picks up `.eval.js` files, so this will need to be converted to the new format to be executed.

test/evals/scenarios/service-usage-disabled.js
```diff
@@ -0,0 +1,25 @@
+/*
```
dylan-klein, line 1: This scenario file is already included in #181, so it can be removed after a rebase.

test/evals/run.js
```diff
@@ -218,6 +218,7 @@ async function main() {
 
     try {
       harness = await createIntegrationHarness({
+        backend,
```
dylan-klein, line 221: Note: #181 also modifies this harness creation logic, so this will need a quick reconciliation.

dylan-klein (approved):
Approve with a few heads-up items for alignment with #181.

PR #154, opened 2026-05-05
chore(evals): commit initial baseline; turn drift detection on

timfee:
Closes #148 

Turns drift detection on by checking in `test/evals/baseline.json`, captured from the first scheduled-style workflow run on `main` ([run 25407323505](https://github.com/google/chrome-enterprise-premium-mcp/actions/runs/25407323505)). Also moves the diff script next to the rest of the eval module so the whole module sits in one place.

## Changes

- **`test/evals/baseline.json`** — first cron-style run on `main`, kept as-is (full per-case responses included).
- **`scripts/eval-diff.js` → `test/evals/diff.js`** — the script is purely an eval concern and now sits next to `test/evals/run.js` and `test/evals/analyze-flaky.js`, both of which are top-level CLI entry points for the eval module. The `scripts/` directory existed only for this one file and is removed.
- **`.github/workflows/agent-evals.yml`** — `hashFiles()` guard and diff invocation point at the new baseline and diff paths.
- **`.gitignore`** — unchanged from `main` (the earlier carve-out to track a file under `results/` is no longer needed since the baseline lives outside `results/`).

## Baseline

- 77 of 92 runs passed (**83.7% pass rate**)
- 15 failed
- 0 transient errors
- INCONCLUSIVE: false

| Category | Pass rate | Counts |
|---|---|---|
| connectors | 100% | 12/12 |
| knowledge | 90.9% | 30/33 |
| boundary | 83.3% | 5/6 |
| inspection | 81.8% | 9/11 |
| discovery | 80% | 4/5 |
| prompt | 73.7% | 14/19 |
| licensing | 50% | 1/2 |
| mutation | 50% | 2/4 |

After this lands, every cron + `workflow_dispatch` on `main` runs `node test/evals/diff.js test/evals/baseline.json results/eval-latest.json`. A drop of more than 5% opens or updates the `Eval drift detected` issue (label `auto:evals`); INCONCLUSIVE runs (transient share above 20%) skip the comparison.

## How to update the baseline later

When an intentional improvement bumps the pass rate, regenerate from a fresh artifact:

```
gh run download <RUN_ID> --repo google/chrome-enterprise-premium-mcp --dir /tmp/eval-artifact
cp /tmp/eval-artifact/eval-results/eval-latest.json test/evals/baseline.json
```

Then commit. The diff script picks up the new figure on the next run.

## Test plan

- [x] `jq '.summary' test/evals/baseline.json` reports `passRate: 83.7`, `inconclusive: false`.
[...]

PR #181, opened 2026-05-08
refactor(evals): migrate cases to eval-as-code

### LINKED ISSUE #84: RFC: Eval formatting
## The current format

The eval cases are stored as markdown files under `test/evals/cases/`.

**File and case structure:**

- Each file holds one or more cases, separated by a literal `--- CASE ---` line.
- Each case has a YAML frontmatter block (id, category, tags, scenario or fixtures, expected_tools, priority).
- Each case has three markdown sections: `## Prompt`, `## Golden Response`, and (optionally) `## Judge Instructions`.
- The loader extracts the sections with a regex against the heading line.

**Execution:**

- The runner applies the case's scenario or fixtures to the in-memory fake API server.
- It drives the agent against the prompt.
- It runs deterministic assertions (forbidden patterns, required patterns, `expected_tools`).
- It calls the LLM judge against the golden response.

The suite has grown to 79 cases across 25 files, and contributors hit five recurring problems when editing it.

### Two ID conventions are in use

Most files use short opaque codes such as `i02`, `m03`, `pr07`, and `b04`. Six cases in `test/evals/cases/tools/list_dlp_rules.md` instead use long descriptive IDs like `list_dlp_rules__when_no_rules_returned_agent_reports_none`. Nothing documents a preference, so new cases inherit whichever convention their file already uses.

### The README's category list is out of date

The README documents `knowledge`, `inspection`, `troubleshooting`, `mutation`, and `discovery`. The cases actually use:

- `knowledge` (33)
- `prompt` (19)
- `inspection` (11)
- `boundary` (6)
- `discovery` (5)
- `mutation` (3)
- `licensing` (2)

Three categories present in the cases (`prompt`, `boundary`, `licensing`) are missing from the README. One category in the README (`troubleshooting`) has no cases.

### Small cases are mostly scaffolding

The simplest case in the suite is a five-word prompt, an eleven-word golden response, and a one-sentence rubric, wrapped in 22 lines of YAML, delimiters, and section headings:

```markdown
--- CASE ---

id: list_dlp_rules__when_no_rules_returned_agent_reports_none
category: discovery
tags:
  - tools
  - dlp
  - empty-state
scenario: no-dlp-rules
expected_tools:
  - list_dlp_rules
priority: P2

## Prompt

List my configured Chrome DLP rules.

## Golden Response

No Chrome DLP rules were found in this organization.

## Judge Instructions

The agent must clearly and accurately state that no rules were found.
```

### The loader misses parser edge cases

The `--- CASE ---` delimiter is a literal text match. The section extractor is the regex `^##\s+<heading>\s*$`. Three failure modes get past the loader without producing any error:

- When a `## Heading` line ends with a trailing space, the loader drops that section's content.
- When a `--- CASE ---` delimiter has a typo, the loader merges the two adjacent cases.
- When a case is missing the `## Prompt` heading, the loader sets `prompt: null` and the case passes every check until the agent call eventually fails.

Sub-issues #48 and #49 are recent reports of related runner edge cases.

### Bundling has no organizing principle

Each of the 25 files contains 2 to 12 cases. During the PR #16 review, Shantanu caught an `m02` DLP-rule-creation case that a contributor had written into `enable_chrome_enterprise_connectors.md` despite being a DLP-rule case. With 25 mixed-purpose files, a reader scanning the directory cannot tell from the filenames what each file actually covers.

## Options considered

Each option below is shown with the same two real cases translated into it:

- The minimal `list_dlp_rules` empty-state case quoted above.
- The richer `pr13` case from `test/evals/cases/prompts/optimize.md`.

### A. Adopt Promptfoo

[Promptfoo](https://www.promptfoo.dev/) is the dominant Node/TypeScript declarative-eval framework.

What it provides:

- YAML test files
- Pluggable assertions through a `type:` discriminator
- Built-in `llm-rubric` for LLM-as-judge
- Built-in `trajectory:tool-used` and `trajectory:tool-args-match` for tool-call assertions
- Reporting, sharding, a web UI, caching, and concurrency
- Provider abstractions for OpenAI, Anthropic, and Vertex; custom providers as JavaScript modules

The two cases in Promptfoo's format:

```yaml
# test/evals/promptfooconfig.yaml
prompts:
  - file://prompts/inline.txt
providers:
  - file://providers/cep-mcp.js   # custom provider that talks to the MCP server

defaultTest:
  assert:
    - type: llm-rubric
      provider: google:gemini-3.1-flash-lite-preview
      value: |
        You are an expert evaluator for an enterprise security AI agent.
        Evaluate strictly but prioritize technical correctness over wording.

tests:
  - description: list_dlp_rules — empty state
    vars:
      prompt: List my configured Chrome DLP rules.
      scenario: no-dlp-rules
    metadata: { category: discovery, priority: P2, tags: [tools, dlp, empty-state] }
    assert:
      - type: trajectory:tool-used
        value: list_dlp_rules
      - type: llm-rubric
        value: The agent must clearly and accurately state that no rules were found.

  - description: optimize — high-noise rule (pr13)
    vars:
      promptName: cep:optimize
      scenario: high-noise-rule
    metadata: { category: prompt, priority: P2, tags: [prompt, noise] }
    assert:
      - type: trajectory:tool-used
        value: diagnose_environment
      - type: trajectory:tool-used
        value: get_chrome_activity_log
      - type: llm-rubric
        value: |
          The agent MUST identify the audit rule as the highest-noise rule by
          analyzing the event log. If the agent reports all rules as equally
          noisy or fails to correlate events to specific rules, grade as FAIL.
          Concrete recommendations for reducing noise are required for PASS.
```

What it fixes:

- Promptfoo owns the parser, so the in-house markdown parser is no longer maintained.
- Promptfoo doesn't have an ID rule or a category set to enforce, so the drift problems are gone — by removing the constraints rather than fixing them.

What it doesn't fix:

- Promptfoo's `tests:` array is multi-case by default. The bundling problem remains.

What it costs:

- A contributor has to write a custom Promptfoo provider in JavaScript that re-implements the scenario lifecycle, the MCP transport, and the tool-trajectory capture against Promptfoo's interfaces.
- Per-case judge rubrics no longer have a dedicated section. A contributor writes the rubric inline under each test.

### B. JSONL dataset

[Inspect](https://inspect.aisi.org.uk/)'s `Sample` shape, and the equivalent shape OpenAI Evals uses, is the most common pattern in research evals: one JSON object per line with fields `id`, `input`, `target`, and `metadata`. Every language has a JSONL parser.

The two cases as JSONL:

```jsonl
{"id":"list_dlp_rules__no_rules","input":"List my configured Chrome DLP rules.","target":"No Chrome DLP rules were found in this organization.","metadata":{"category":"discovery","priority":"P2","tags":["tools","dlp","empty-state"],"scenario":"no-dlp-rules","expected_tools":["list_dlp_rules"],"judge_rubric":"The agent must clearly and accurately state that no rules were found."}}
{"id":"pr13","input":{"promptName":"cep:optimize"},"target":"Agent should identify that the \"Audit pastes to generative AI sites\" rule is generating significantly more events than other rules (24 events vs 1 from the block rule)...","metadata":{"category":"prompt","priority":"P2","tags":["prompt","noise"],"scenario":"high-noise-rule","expected_tools":["diagnose_environment","get_chrome_activity_log"],"judge_rubric":"The agent MUST identify the audit rule as the highest-noise rule..."}}
```

What it fixes:

- The loader runs every case through a JSON Schema validator and rejects any that breaks the ID rule or the category set.
- Each case is one self-contained record. JSONL has no concept of a multi-case file.
- The loader is `JSON.parse` per line. The current bespoke parser is no longer maintained.

What it costs:

- A contributor writing a multi-paragraph judge rubric inside a JSON string has to escape every quote and newline. There is no real multi-line.
- A reviewer cannot see whitespace bugs in code review.
- Contributors edit rubrics often, because the rubrics are what give the suite its diagnostic value. JSONL makes that path the worst.

### C. Tighten the in-house format

Replace the markdown wrapper and the multi-case bundling with one YAML file per case, organized into per-category subdirectories, validated against a published JSON Schema. The runner, the judge, the scenarios, and the deterministic checks all stay unchanged.

The two cases in this format:

```yaml
# test/evals/cases/discovery/list_dlp_rules__no_rules.yaml
id: list_dlp_rules__no_rules
priority: P2
tags: [tools, dlp, empty-state]
scenario: no-dlp-rules
expected_tools: [list_dlp_rules]

prompt: List my configured Chrome DLP rules.

golden_response: No Chrome DLP rules were found in this organization.

judge_rubric: The agent must clearly and accurately state that no rules were found.
```

```yaml
# test/evals/cases/prompts/pr13_high_noise_rule.yaml
id: pr13
priority: P2
tags: [prompt, noise]
scenario: high-noise-rule
prompt_from: cep:optimize
expected_tools:
  - diagnose_environment
  - get_chrome_activity_log

golden_response: |
  Agent should identify that the "Audit pastes to generative AI sites"
  rule is generating significantly more events than other rules
  (24 events vs 1 from the block rule). Multiple users across multiple
  days are triggering it. Agent should flag this as a high-noise rule
  and recommend either tightening the rule's conditions, switching to a
  more targeted URL list, or accepting the volume if monitoring GenAI
  usage is a priority.

judge_rubric: |
  The agent MUST identify the audit rule as the highest-noise rule by
  analyzing the event log. If the agent reports all rules as equally
  noisy or fails to correlate events to specific rules, grade as FAIL.
  Concrete recommendations for reducing noise (narrowing conditions,
  refining URL matching, or accepting noise with justification) are
  required for PASS.
```

#### One YAML file per case

Drop the markdown wrapper, the `--- CASE ---` delimiter, and the `## Prompt` / `## Golden Response` / `## Judge Instructions` section markers. Each frontmatter field and each prose section becomes a YAML key. YAML block scalars (`|`) handle multi-line prose without escaping.

A contributor writes the minimal case in 9 lines instead of 22.

#### Category comes from the directory

Each case file's path is `test/evals/cases/<category>/<case-id>.yaml`. The directory name is the category — no separate `category` field. A contributor adds a category by creating a directory and renames one by moving files. The README documents whichever directories exist.

#### JSON Schema validates every case at load time

The schema enforces:

- A single ID rule across the suite (one regex)
- The valid category set
- Required and optional fields per case, with type checks
- Constraints on `expected_tools`, `forbidden_patterns`, etc.

A schema violation is a file-level error and stops the run before any case executes.

#### Smaller loader

The loader is `js-yaml` plus a JSON Schema check. Both are well-maintained upstream. The current ~200-line custom loader and the section-extraction regex are no longer needed.

#### What stays unchanged

- The runner (`test/evals/run.js`)
- The judge prompt and rubric (`test/evals/lib/judge.js`)
- Every scenario file under `test/evals/scenarios/`
- The fake API server and its `setState` lifecycle
- The deterministic forbidden/required-pattern checks
- The `expected_tools` strict-presence check

#### Cost

The repository has 79 separate files instead of 25, organized into per-category subdirectories.

### Why not A

A contributor switching the suite to Promptfoo would:

- **Delete** about 200 lines of loader plus a few dozen for the section parser.
- **Write** a custom Promptfoo provider that re-implements the scenario lifecycle, the MCP transport wiring, and the tool-trajectory capture against Promptfoo's interfaces.

The new provider is roughly the same size as the deleted loader, so the line count of bespoke code barely moves. The new code is also integration glue between two upstream frameworks. Integration glue is historically the most expensive code to maintain, because an upstream change in either framework can break it.

### Why not B

JSONL handles the parser and the bundling cleanly. The cost is on the editing path:

- A contributor writing a multi-paragraph judge rubric inside a JSON string has to escape every quote and newline.
- A reviewer cannot see whitespace bugs in code review.
- Contributors edit rubrics often, because the rubrics are what give the suite its diagnostic value.

Optimizing the parsing path at the cost of the editing path is the wrong trade for this codebase.

### Why C

- Every problem listed above is addressed.
- The runner and the judge are untouched.
- The schema enforcement is JSON Schema, a standard tool with well-understood error reporting.
- The format is still in-house, but the surface area is much smaller.
- A contributor can script most of the migration from the existing markdown. (Migration is out of scope for this RFC.)

## Open questions

1. **Where does `category` live — in the directory or in a frontmatter field?** Directory inference is simpler and removes the field-vs-directory drift; the cost is that renaming a category turns into a directory move.

2. **Should the format let multiple cases share a single judge rubric?** Several cases share boilerplate phrasing. YAML anchors or a `!include` extension would let a case reuse a shared rubric, at the cost of a YAML construct contributors have to learn.

3. **Should `expected_tools` support `forbidden` or `exact-match` modes?** The runner today checks every named tool was called. The forbidden-tool and exactly-these-tools cases haven't come up. If either matters, the schema should encode it now rather than retrofit it later.

timfee:
96 markdown eval cases plus 2 cherry-picked from #180 become 99 `.eval.js` files. They live under `test/evals/cases/<category>/<id>-<slug>.eval.js`. Resolves #84.

## Changes

- **Loader (`test/evals/lib/loader.js`).**
  - Markdown parser and section-extraction regex deleted.
  - New path: glob `**/*.eval.js`, `import()` each.
  - Inline validation:
    - id pattern
    - allowed fields
    - prompt-or-promptName required
    - scenario + fixtures mutex
    - duplicate-id detection across the suite
- **Cases (`test/evals/cases/`).**
  - 99 `.eval.js` files in 10 per-category directories.
  - Zero `.md` files remain.
- **IDs normalized.** 19 long-descriptive IDs become `<prefix><NN>`:
  - `get_connector_policy__*` → `c01..c12`
  - `list_dlp_rules__*` (inspection) → `i09..i11`
  - `list_dlp_rules__*` (discovery) → `d03..d05`
  - `enable_chrome_enterprise_connectors__*` → `m04`
  - `sys-01` / `sys-02` → `s01` / `s02`
  - `sec-log-01` / `sec-log-02` → `sec01` / `sec02`
  - All other IDs unchanged.
- **Filenames** follow `<id>-<slug>.eval.js`.
  - Slug is a directory-listing aid.
  - Loader doesn't read it.
- **Spec:** `docs/evals/format.md`.
- **Starter:** `test/evals/TEMPLATE.eval.js`.
- **README** rewritten.
  - Old YAML-frontmatter section gone.
  - Replaced with a JS example and a pointer to the spec.
- **Loader tests.**
  - Markdown-loader tests removed alongside the parser.
  - New `loadEvalAsCodeFromFile` tests cover:
    - valid case
    - unknown field
    - invalid id
    - missing prompt/promptName
    - scenario + fixtures mutex
- **PR #180 cases (s04, s05).**
  - Cherry-picked here in the new format.
  - Their two scenario files come along.
  - `run.js` harness wiring and fake-server 403 logic stay on #180.

## Commits

13 commits.

- One per category (knowledge, prompt, connectors, inspection, boundary, discovery, mutation+licensing+system+security combined).
- Three infrastructure commits (loader, experiments-field fix, docs+template+README).
- One filename-normalization commit (86 git mvs).
- One s04/s05 port commit.
- One cleanup commit (markdown loader removal).

Every commit has the suite passing `--dry-run` end-to-end.

## Coordination

- **#180 (`feat/api-enablement-remediation`, approved).**
  - After this lands: rebase against main.
  - Delete the `.md` versions of s04/s05.
  - Keep the `run.js` and fake-server changes.
- **#154 (`chore/commit-eval-baseline`).**
  - After this lands: rebase.
  - Recapture `baseline.json` against the migrated suite (99 cases, new IDs).

## Test plan

- [x] `node --test test/local/eval-loader.test.js` — 13/13 pass.
- [x] `npm run eval -- --dry-run` — 99/99 pass.
- [x] `npm run eval --output current.json` (Gemini judge, fake backend, single run) — 81/99 pass.
- [x] Diff vs. #154 baseline (rename map applied):
  - 70 stable PASS→PASS.
  - 9 stable FAIL→FAIL.
  - 5 first-run regressions: all rerun-PASS (LLM noise).
  - 1 flaky pre-existing (`i02`, 1/3 in stress test).
  - 1 model drift on main (`m03`, 0/3 on both branches).
  - Migration-caused regressions: 0.
- [ ] Reviewer spot-checks migrated cases against original markdown for prose fidelity.

## Resolves #84

Updated comment on the issue covers shape, diff vs. Option C, ID normalization, and the validation result.


timfee:
@dylan-klein tried my hand at the consistent eval file format. Nothing urgent, but an implementation of the RFC for #84

dylan-klein:
I noticed that many of the migrated cases now use numbered IDs (e.g., k01, d02) instead of descriptive snake_case names. Could you clarify the reasoning behind this shift? Personally, I find descriptive IDs more helpful for identifying the purpose of a case at a glance.

dylan-klein (approved):

timfee:
> I noticed that many of the migrated cases now use numbered IDs (e.g., k01, d02) instead of descriptive snake_case names. Could you clarify the reasoning behind this shift? Personally, I find descriptive IDs more helpful for identifying the purpose of a case at a glance.


Ah, yes. I updated the titles of the files but not the ID. It seems like it should extract the nuneric ID and slug from filename vs duplicating it (or i could write a lint rule that forces them to match), or something else if you had a preference :)

dylan-klein:
> > I noticed that many of the migrated cases now use numbered IDs (e.g., k01, d02) instead of descriptive snake_case names. Could you clarify the reasoning behind this shift? Personally, I find descriptive IDs more helpful for identifying the purpose of a case at a glance.
> 
> Ah, yes. I updated the titles of the files but not the ID. It seems like it should extract the nuneric ID and slug from filename vs duplicating it (or i could write a lint rule that forces them to match), or something else if you had a preference :)

I personally would rather move in the other direction and have all tests have descriptive snake_case names. I feel having a very succinct description of what a test is doing is a good practice. That can be done in a later dedicated CL though.

TusharKanekiDey (approved):

Issue #183, opened 2026-05-08
refactor: Address DRY violations and streamline API client architecture


timfee:
# Refactor: Address DRY violations and streamline API client architecture

## Motivation
A static analysis run of `jscpd` detected several code duplication (DRY) violations in the Model Context Protocol (MCP) server:
1.  **DLP Rule Parsing**: `get_dlp_rule` and `list_dlp_rules` tools had duplicated logic for formatting rule displayNames, state Title Cases, action mapping, and cleaning up triggers.
2.  **ADC Probing**: `mcp-server.js` had a duplicate local implementation of Application Default Credentials (ADC) probing (`probeADC`) that was almost identical to the standard `adcCredential().probe()` factory in `lib/util/credential/adc.js`.
3.  **Unused Code**: `lib/util/google-auth-provider.js` was completely unused and duplicated standard auth setup.
4.  **Knowledge generation**: `11-dlp-rule-reference.doc.js` duplicated about 40 lines of markdown generation logic that was already present in constants.
5.  **Internal API Client Logic**: Consolidate policy retrieval in `RealCloudIdentityClient` and licensing 403 error handling in `RealAdminSdkClient`.

Additionally, the architectural pattern of having an `interfaces/` abstract directory and `real_*.js` production client prefix is redundant in JavaScript. In this codebase, we do not use client-side mock classes (e.g., `FakeCloudIdentityClient`). Instead, integration tests use the *production* client pointed to an in-process fake HTTP Express server (`fake-api-server.js`) by overriding the `rootUrl` option.
Therefore, the JSDoc interface classes were pure overhead.

---

### PR 1: DRY Improvements & Cleanup
*   **Commit 1**: Extract `parseDlpRule(policy)` helper in `lib/util/helpers.js` using robust trigger cleaning. Refactor `get_dlp_rule.js` and `list_dlp_rules.js` to use it.
*   **Commit 2**: Update `CredentialProbe` to include `quotaProject` and resolve it in `adc.js`. Migrate `mcp-server.js` to use `adcCredential().probe()` and remove `probeADC`.
*   **Commit 3**: Delete unused `google-auth-provider.js` and clean up references in labeler, expected-tarball, and README.
*   **Commit 4**: Align `generateDlpCelReference` and refactor `11-dlp-rule-reference.doc.js` to consume it.
[...]

PR #190, opened 2026-05-08
refactor(auth): thread credential source for accurate remediation advice

dylan-klein:
# Pull Request #190 Description

## 🎯 Motivation
Currently on `main` (HEAD), the server returns a single, generic set of terminal re-authentication instructions (`chrome-enterprise-premium auth login`) for all authentication failures. This generic advice creates unnecessary friction depending on the runtime environment:
*   **Gemini CLI Users** are forced to switch contexts to a terminal shell, when they could easily re-authenticate inline via `/mcp reauth`.
*   **Hosted Service Accounts** (using `GOOGLE_APPLICATION_CREDENTIALS`) are instructed to run interactive CLI logins, which is impossible in non-interactive container environments (where operators need Cloud API/IAM verification steps instead).

This PR introduces **Environment-Aware Remediation** by threading the resolved credentials source (`bearer`, `cache`, `adc` [Service Accounts], or `provided` [custom clients]) from resolution down to the error handler, tailoring remediation advice to the specific environment. It also shiores up mock server security by resolving CodeQL Prototype Pollution vulnerabilities in the test harness.

## 🛠️ Main Changes
*   **Credentials Source Threading**: Propagates the resolved `credentialsSource` through the entire API client instantiation pipeline to the tool execution wrapper.
*   **Context-Aware Remediation**: Tailors re-auth instructions based on the source:
    *   `bearer` (Gemini CLI): Instructs the user to run `/mcp reauth in your Gemini CLI`.
    *   `cache` (Local Dev): Instructs the user to run `mcp login in your terminal`.
    *   `adc` (Service Accounts): Directs the user to verify GCP Cloud APIs and IAM permissions (since interactive shells are unavailable).
    *   `provided` (Custom Client): Instructs the user to validate their caller-supplied client configuration.
*   **Hardened Mock Server Security**: Patched dynamic route lookups and fixture ingestion in the fake Google API mock server (`test/helpers/fake-api-server.js`) using `isSafeKey` checks, eliminating CodeQL Prototype-polluting assignment vulnerabilities.
*   **DRY Remediation Mapping**: Consolidated commands and session context messaging into a clean `REAUTH_ADVICE` lookup table.
*   **Test Isolation & Mocks**: Introduced deterministic unit tests covering the throw-and-fall-through path for expired credentials and decoupled local machine state in BDD suites.

## 💡 Non-Trivial Design Decisions & Validation
*   **Preserving DWD and SA Impersonation**: Retained manual JWT service-account key file loading rather than raw GoogleAuth configuration to ensure full support for domain-wide delegation (`CEP_IMPERSONATE_SUBJECT`).
*   **Fail-Fast Integration**: Combined the new asynchronous pre-flight validity checks (`isTokenLocallyValid`) with the PR's `credentialsSource` extraction. If a cached token is missing or invalid, it fails fast immediately without initiating unnecessary backend or wrapper logic.
*   **Rate Limiting Exemption**: Intentionally bypassed and dismissed application-level rate limiting on HTTP routes, keeping it cleanly delegated to the hosting infrastructure (Cloud Run / GCLB / reverse proxies) as documented in the architecture guidelines.

## 🧪 Testing Verified
*   **Lint & Formatting**: Verified style guidelines and Prettier compliance via `npm run lint`. (Passed)
*   **Unit Tests**: Verified 100% coverage across all **481 unit tests** via `npm run test:unit`. (Passed)
*   **Integration Tests**: Verified full end-to-end mock backend logic via `npm run test:integration:fake`. (Passed)


timfee (commented):
Duplicate review — see [pullrequestreview-4256208838](https://github.com/google/chrome-enterprise-premium-mcp/pull/190#pullrequestreview-4256208838) for the actual content.

tools/utils/wrapper.js
```diff
@@ -251,8 +261,8 @@ export function guardedToolCall(
           errorMessage.includes('invalid_grant')
             ? 401
             : 403)
-        const isOAuth = !!context?.authToken || !!context?.requestInfo?.headers?.authorization
-        const remediationMessage = getAuthRemediationMessage(resolvedStatus, isOAuth)
+        const source = error.credentialsSource || (authToken ? 'bearer' : 'adc')
```
timfee: If `error.credentialsSource` is unset, `source` is `'adc'` whenever `authToken` is falsy. That covers every error path that doesn't pass through `handleApiError` — validation throws, SDK rejections that bypass the helper, retry-helper throws. A `mcp login` user gets gcloud-flavored remediation when this happens. That's the failure mode named in the PR description.

Fix: stash `source` on the `context` object the wrapper passes to `handler` at line ~213, after `getAuthClient` returns. Then `context.credentialsSource` is the primary read here. `error.credentialsSource` is the secondary. The ADC default is unreachable when `handleApiError` is bypassed.

lib/util/auth.js
```diff
@@ -26,26 +26,32 @@ import { getAuthErrorMessage } from './auth-error.js'
...
   }
 
   const auth = new GoogleAuth({ scopes })
   try {
-    return await auth.getClient()
+    const client = await auth.getClient()
+    // Verify the client is functional before returning it.
+    // This catches cases where ADC is "configured" but invalid/expired
+    // (e.g. gcloud credentials that require reauth).
+    await client.getAccessToken()
```
timfee: The `catch (adcError)` block is reachable only when `await client.getAccessToken()` throws. That's the whole reason this line was added. No test covers it: `auth.test.js` stubs `getAccessToken` to succeed, and `auth-oauth-cache-fallback.test.js` covers the case where `getClient` itself throws.

Add a test where `GoogleAuth.prototype.getClient` returns a client whose `getAccessToken` rejects, and assert the result has `source === 'cache'`. Without that, removing `await client.getAccessToken()` in a future refactor passes CI.

lib/util/api-client.js
```diff
@@ -27,13 +27,16 @@ import { getAuthClient } from './auth.js'
  * @param {string[]} scopes - OAuth scopes required.
  * @param {string} [authToken] - Optional auth token.
  * @param {object} [apiOptions] - Additional options (e.g. rootUrl).
- * @returns {Promise<object>} The API service client instance.
+ * @returns {Promise<{client: object, source: string}>} An object containing the API service client instance and its credential source.
  * @throws {Error} If authentication or client creation fails.
  */
 export async function createApiClient(service, version, scopes, authToken, apiOptions = {}) {
   let authClient = apiOptions.auth
+  let source = 'provided'
```
timfee, line 35: When the caller pre-supplies `apiOptions.auth`, `source` is `'provided'`. `getAuthRemediationMessage` has no `'provided'` arm — only `bearer`, `cache`, and a default that is the gcloud-ADC copy. A 401 from a caller-supplied client therefore returns the gcloud message even when the caller's actual flow has nothing to do with ADC.

Two options: leave `source` undefined here, so `(authToken ? 'bearer' : 'adc')` in the wrapper is the source instead. Or add a `'provided'` arm in `getAuthRemediationMessage` with a generic "auth failed; check the credentials your caller passed in" message. Either is fine. Current behavior is just the wrong default for that path.

tools/utils/wrapper.js
```diff
@@ -39,6 +39,16 @@ function getAuthRemediationMessage(status, isOAuth = false) {
 2. **Verify APIs are enabled:** Ensure \`admin.googleapis.com\`, \`chromemanagement.googleapis.com\`, \`chromepolicy.googleapis.com\`, and \`cloudidentity.googleapis.com\` are enabled in your Google Cloud project.`
   }
 
+  if (source === 'cache') {
```
timfee: This `cache` block and the `bearer` block at line 32 differ only in the reauth command (`mcp login` vs `/mcp reauth`) and one sentence of explanatory text. A third source means a third near-duplicate.

Could be a small lookup:

```js
const REAUTH = {
  bearer: { cmd: '/mcp reauth in your Gemini CLI', sessionLabel: 'OAuth session' },
  cache:  { cmd: 'mcp login in your terminal',     sessionLabel: 'cached tokens' },
}
```

Then build the 401/403 strings from the entry. Not blocking; skip if you'd rather keep the explicit shape for grep-ability.

timfee (commented):
Quick sanity check, four inline comments. None of these block merge on their own.

Bigger thing first: this branch is five merges behind `main`. PR #183 (`refactor: address DRY violations and streamline API client architecture`) renamed `lib/api/real_*_client.js` to `lib/api/*_client.js` and removed `lib/api/interfaces/`. Every API-client file edited here has either moved or been deleted on `main`. Rebase first; the diff against current `main` is mostly rename conflicts.

The four inline comments: cache-auth misclassified as ADC in the wrapper fallback; the new `getAccessToken` validation has no test for the throw-and-fall-through path; `source: 'provided'` is unhandled downstream; bearer and cache remediation copy is a near-duplicate.

tools/utils/wrapper.js
dylan-klein: Fixed by resolving the credential source once in the tool wrapper and passing it through to the error handler.

lib/util/auth.js
dylan-klein: Added a unit test in `test/local/auth-oauth-cache-fallback.test.js` covering the throw-and-fall-through path.

lib/util/api-client.js
dylan-klein: Added a `'provided'` arm to `getAuthRemediationMessage` with generic instructions for caller-supplied clients.

tools/utils/wrapper.js
dylan-klein: Refactored remediation messages to use a lookup table (`REAUTH_ADVICE`) for commands and labels.

PR #210, opened 2026-05-14
fix(evals): update agent model and add catastrophic-failure alarm

timfee:
## Summary

- Repoint eval agent model from retired `gemini-2.5-flash` to `gemini-3.1-flash-lite-preview` (the model the judge already uses)
- Classify non-retryable agent errors as TRANSIENT so a fully broken run is INCONCLUSIVE, not a noisy drift regression
- Add a separate catastrophic-failure detector and a separate "Eval infrastructure failure" issue stream, with the dominant error message in the body

## Why

The Agent Evals cron has been red since 2026-05-09. Every case 404'd against `gemini-2.5-flash`, which is no longer in the v1beta ListModels response. The runner classified each 404 as FAIL (the `catch` block at `run.js:274-276` wrote the error string to `responseText`; the judge then judged the error string and predictably failed it). FAIL is in the pass-rate denominator, so the run was indistinguishable from a real regression in the diff even though the model never ran. The drift-issue comments contained only `0% pass rate`, a list of every case ID, and no error message context.

Two structural problems followed:

1. Without a fix, the same noise recurs on the next model-id removal. The fix is to bucket non-retryable agent errors as TRANSIENT so the existing INCONCLUSIVE ceiling (>20% transient -> skip diff) handles them.
2. Even with INCONCLUSIVE skipping the drift comparison, a fully broken run is silent. The fix is a separate "catastrophic" detector that exits non-zero on `passed === 0`, with its own issue stream and the dominant error message in the body.

## Test plan

- [x] `node test/evals/infra-check.js .../eval-latest.json` exits 1 with a usable summary on the current broken artifact from run #25797101722 (`(101x) Agent error: ... 404 Not Found ...`)
- [x] `node --check` passes on all changed JS
- [x] `node test/evals/run.js --dry-run --id b01` still loads the eval config
- [ ] After merge: manual `workflow_dispatch` run against `main` to confirm green end-to-end
- [ ] Existing drift issue #194 can be closed once the next scheduled run succeeds

### DIFF
--- a/.github/workflows/agent-evals.yml
+++ b/.github/workflows/agent-evals.yml
@@ -48,6 +48,20 @@ jobs:
           path: results/
           retention-days: 30
 
+      # Catastrophic-failure check is independent of drift comparison. The
+      # trigger is zero passing cases. The most likely cause for that is
+      # infrastructure failure (model id removed upstream, expired key, proxy
+      # outage), not a real regression. Runs whether or not a baseline exists.
+      - name: Check for catastrophic failure
+        id: infra
+        if: hashFiles('results/eval-latest.json') != ''
+        run: |
+          set +e
+          node test/evals/infra-check.js results/eval-latest.json \
+            | tee infra-output.txt \
+            | tee -a "$GITHUB_STEP_SUMMARY"
+          echo "exit_code=${PIPESTATUS[0]}" >> "$GITHUB_OUTPUT"
+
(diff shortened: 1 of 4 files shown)


dylan-klein (approved):
LGTM. The model update and infrastructure failure detection are solid improvements to the eval suite stability.

PR #212, opened 2026-05-14
test: add exact version matching check for gemini-extension.json

TusharKanekiDey:
Adds a strict version matching assertion between `package.json` and `gemini-extension.json` to ensure the extension metadata stays perfectly in sync.


timfee (approved):
LGTM as a failsafe-- hopefully this never triggers as CI handles keeping these in sync :)

PR #213, opened 2026-05-14
ci: specify container_properties for Kokoro Korbe V2 build execution

TusharKanekiDey:
Resolves the Kokoro fatal ConfigException (`container_properties must be set in the build config`) by explicitly specifying the Kokoro Ubuntu 20.04 container image in `presubmit.cfg` and `continuous.cfg`.


dylan-klein (approved):
Approving: The Kokoro container fix is required for CI, and the baseline.json formatting is necessary for the presubmit check to pass.

PR #221, opened 2026-05-15
fix(ci): install from public npm registry; only use Wombat for publish

timfee:
## Summary
The npm-publish job set Wombat Dressing Room as the registry for the entire job, which made \`npm ci\` route through Wombat and 404 on packages that aren't allowlisted there (e.g. \`zod-to-json-schema@3.25.2\`). Latest failure: https://github.com/google/chrome-enterprise-premium-mcp/actions/runs/25893075172/job/76100257998

Splits into two \`setup-node\` invocations: install from the public npm registry, then re-set the registry to Wombat for the publish step (where the auth token applies).

## Test plan
- [ ] Next release triggers npm-publish; \`npm ci\` succeeds and the publish step still authenticates against Wombat.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -10,12 +10,19 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v4
+      # Install from the public npm registry. The Wombat Dressing Room only
+      # proxies packages this org has explicitly allowlisted, so using it for
+      # `npm ci` 404s on common dev deps (e.g. zod-to-json-schema).
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: 'npm'
+      - run: npm ci
+      # Re-run setup-node to point auth at Wombat for `npm publish` only.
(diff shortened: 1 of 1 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -10,12 +10,19 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v4
+      # Install from the public npm registry. The Wombat Dressing Room only
+      # proxies packages this org has explicitly allowlisted, so using it for
+      # `npm ci` 404s on common dev deps (e.g. zod-to-json-schema).
       - uses: actions/setup-node@v4
         with:
           node-version: '20'
```
TusharKanekiDey: Is it possible to make it 24 or will it start breaking the flow ?

TusharKanekiDey (approved):

.github/workflows/npm-publish.yml
timfee: Let me try; should be fine!

PR #223, opened 2026-05-15
chore(deps): upgrade zod 3 -> 4; migrate to non-deprecated APIs

timfee:
## Summary
Upgrades `zod` from 3.25.76 to 4.4.3 and migrates all usage to the v4 idiom.

### Source rewrites
- **`z.object({...}).passthrough()` → `z.looseObject({...})`** (47 sites). Per [v4 changelog](https://zod.dev/v4/changelog), the chainable methods are deprecated in favor of top-level constructors.
- **`z.record(<value>)` → `z.record(z.string(), <value>)`** (3 sites). Single-argument `z.record()` is removed in v4; both key and value schemas are required.

### Test updates
- `test/local/output_schema_validation.test.js` was checking the v3 internal `schema._def.unknownKeys === 'passthrough'`. v4 represents loose objects via a catchall — that field no longer exists. Replaced with a behavioral check: parse a probe with an unknown key and assert the key survives.
- `test/integration/tools/connector_enablement.test.js` asserted on v3-specific zod error message strings (`Array must contain at least 1 element(s)`, `Invalid enum value`). v4 emits structured JSON with `"code": "too_small"` / `"code": "invalid_value"`; updated the regexes accordingly.

### Codemod
The mechanical rewrites were applied via a small AST-aware-ish node script that walks balanced parens to find the `z.object(...)` matching each `.passthrough()`. Not committed; the diff is the artifact.

## Test plan
- [x] `npm run lint` clean
- [x] `npm test` (unit + integration + smoke) passes locally
- [ ] CI green


test/local/output_schema_validation.test.js
```diff
@@ -43,24 +43,25 @@ describe('Output Schema Validation', () => {
...
-  test('When tools are registered, then every tool has a valid top-level z.object().passthrough() outputSchema', () => {
+  test('When tools are registered, then every object outputSchema preserves unknown keys', () => {
+    // Behavioral check: parsing a probe with an unknown field should keep that
+    // field. zod v4 represents loose objects via a catchall; checking behavior
+    // is more robust than poking at internals.
+    const PROBE_KEY = '__cep_unknown_probe__'
     for (const [name, tool] of registeredTools) {
       const schema = tool.schema.outputSchema
-      if (!schema) {
+      if (!schema || schema.def?.type !== 'object') {
```
dylan-klein, line 53: ### Technical Regression: False Pass in Validation Loop

The check `schema.def?.type !== 'object'` appears to be a logic bug. In Zod, internal metadata is stored on the `._def` property, and `.def` is `undefined`.

**The Bug:**
Because `schema.def` is `undefined`, the expression `schema.def?.type !== 'object'` evaluates to `true` for every tool, causing the loop to `continue` (skip) every tool in the registry. This test is currently providing a 'False Pass' and is not verifying the architectural requirement.

**Fix:**
Change to `schema._def.typeName === 'ZodObject'` (or check for the existence of the `.parse()` method).

test/integration/tools/connector_enablement.test.js
```diff
@@ -128,7 +128,8 @@ describe('Connector Enablement Integration', () => {
     })
 
     assert.ok(result.isError, 'Tool result should indicate error for empty list')
-    assert.match(result.content[0].text, /Array must contain at least 1 element\(s\)/)
+    assert.match(result.content[0].text, /"code": "too_small"/)
```
dylan-klein: ### Inquiry: Change in External Error Message Format

**Question:**
Is the shift from human-readable strings to structured JSON error codes in the tool's output intended?

**Impact:**
While technically accurate, Zod v4's default `.toString()` output is a JSON array (e.g., `[{"code":"too_small",...}]`). If downstream agents or UIs surface this directly, users will see technical JSON instead of the previous clear instruction ("Array must contain at least 1 element(s)"). We may want to use a custom formatter to maintain the user-facing string quality.

test/local/output_schema_validation.test.js
timfee: In v4, `.def` is the public alias for `_def` — same object. Probed locally:

```
Object.is(schema.def, schema._def) // true
schema.def?.type                    // 'object' for ZodObject
schema._def.typeName                // undefined in v4
```

So `schema.def?.type !== 'object'` is false for every ZodObject and the loop body runs. Confirmed by instrumenting the test against the registered tools:

```
total tools: 22
considered (passed the type check): 22
loose-confirmed (probe survived): 22
```

The behavioral assertion verifies the same property the v3 `_def.unknownKeys === 'passthrough'` check did — that unknown keys survive parse — just via behavior instead of introspection.

The proposed `_def.typeName === 'ZodObject'` would cause the false-pass it warns against: v4 dropped that field, so the comparison would be false for every iteration and the loop would skip everything.

test/integration/tools/connector_enablement.test.js
timfee: Right call — addressed in dd85264. Added `installPrettyValidationErrors()` (`lib/util/mcp_pretty_errors.js`) that wraps `validateToolInput` on the McpServer and routes the ZodError through `z.prettifyError()` before the SDK serializes it. Wired into `getServer()` and the integration harness so both runtime and tests get the same treatment.

Tool callers now see:

```
Invalid arguments for tool enable_chrome_enterprise_connectors:
✖ Too small: expected array to have >=1 items
  → at connectors
```

Updated the assertions here back to the human strings.

dylan-klein (approved):

PR #228, opened 2026-05-15
feat: proactive inline OAuth login and explicit auth tools

dylan-klein:
## Overview
This pull request implements a proactive, inline OAuth login model for the Chrome Enterprise Premium (CEP) MCP server, transitioning away from the previous 'fail on missing/expired auth' behavior. It adopts the robust inline auth pattern implemented in the Google Workspace Extension (gemini-cli-extensions/workspace).

### 1. Truly Automatic Proactive Auth (tools/utils/wrapper.js)
* **Seamless Experience (Desktop)**: When a tool is called without valid credentials, the tool wrapper now blocks, automatically launches the default browser to the Google OAuth consent screen, and waits for the user to sign in. Once complete, the original tool call finishes successfully with the requested data.
* **Agent-Led Headless Support (SSH)**: In headless environments, the server detects the lack of a display and returns a structured "Auth Required" signal. This instructs the agent to invoke the `cep_auth` tool to fetch the consent URL for the user, ensuring a smooth remediation flow without process hangs.
* **Universal Coverage**: This logic is now part of the tool wrapper, ensuring proactive auth works across Stdio, SSE, and Hono transport modes.

### 2. Robust SDK Compatibility & Zod Validation
* **Strict Success, Unstructured Error Model**: Solved a critical issue where the MCP Client SDK would fail to validate "Auth Required" signals against strict tool `outputSchema` definitions.
* **Architectural Choice**: Instead of globally relaxing tool schemas (which would make all data fields optional), we return the auth signal as a standard MCP error (`isError: true`) with unstructured text content.
* **Machine Readability**: We still provide the structured auth payload as a JSON block within the markdown content, allowing agents to reason about the remediation (like the `manualCommand`) while bypassing the rigid SDK type-checker.

### 3. Per-Tool Scope Enforcement & Hardening
* **Granular Scopes**: Every MCP tool now explicitly declares its required Google OAuth scopes.
* **Proactive Gap Detection**: Before a tool call is even attempted, the wrapper checks if the cached token has the required scopes. If not, it proactively triggers the re-authentication flow to escalate permissions.
* **Greedy First-Login**: To minimize user friction, the initial login defaults to requesting all known scopes, ensuring the browser only needs to open once for the entire session.

### 4. Explicit Auth Management Tools (tools/definitions/auth.js)
* Consolidated and registered three standard MCP tools for auth lifecycle management:
  * `cep_auth`: Manages the login session (starts server or completes via pasted URL).
  * `auth_status`: Checks which account is assigned in and what permissions are granted.
  * `auth_clear`: Securely deletes the local token cache (Sign Out).

### 5. Robust Testing & CI Compatibility
* **Zero-Server Test Mode**: Modified the auth flow to skip starting HTTP loopback listeners during unit tests (`CEP_BACKEND=fake`), resolving process hang issues in CI and local test runs.
* **Environment Propagation**: Fixed the unit test runner to correctly pass environment variables to child test processes.

## Verification
* `npm run presubmit` passes 100% clean (ESLint, Prettier, 443 unit tests, 31 integration tests, and smoke tests).
* Verified "Truly Automatic" flow on local Linux workstation (browser pops, tool returns data).
* Verified "Agent-Led" flow via SSH (Agent receives signal, calls `cep_auth`, and presents URL).

### DIFF
--- a/lib/util/auth.js
+++ b/lib/util/auth.js
@@ -29,6 +29,7 @@ limitations under the License.
 import fs from 'node:fs/promises'
 import { JWT, OAuth2Client } from 'google-auth-library'
 import { TokenCache } from './credential/token_cache.js'
+import { SCOPES } from '../constants.js'
 
 /**
  * Module-level cache of JWT credentials, keyed on the inputs that affect
@@ -135,7 +136,7 @@ export async function getAuthClient(scopes, authToken, apiOptions = {}) {
         if (apiOptions.onStatusUpdate) {
           apiOptions.onStatusUpdate('Authentication required. Opening browser for Google OAuth consent...')
         }
-        const result = await startToolAuth({ scopes })
+        const result = await startToolAuth({ scopes: Object.values(SCOPES) })
         if (result.status === 'completed') {
           if (apiOptions.onStatusUpdate) {
             apiOptions.onStatusUpdate('Authorization successful. Tokens cached.')
--- a/lib/util/auth.js
+++ b/lib/util/auth.js
@@ -97,10 +97,11 @@ async function jwtFromServiceAccountFile(keyPath, scopes) {
  * Returns an authenticated client for Google API calls.
  * @param {string[]} scopes - Scopes the client needs (used by the SA-key path).
  * @param {string} [authToken] - Optional pre-acquired bearer token.
+ * @param {object} [apiOptions] - Optional options containing status logging callbacks.
  * @returns {Promise<import('google-auth-library').AuthClient>} An authenticated client.
  * @throws {Error} If no credential source resolves.
  */
-export async function getAuthClient(scopes, authToken) {
+export async function getAuthClient(scopes, authToken, apiOptions = {}) {
   if (authToken) {
     const auth = new OAuth2Client()
     auth.setCredentials({ access_token: authToken })
@@ -113,14 +114,47 @@ export async function getAuthClient(scopes, authToken) {
   }
 
   const cache = new TokenCache(TokenCache.defaultPath())
-  const tokens = await cache.readEnforcingMode()
+  let tokens = await cache.readEnforcingMode()
+
+  if (tokens) {
+    const granted = new Set((tokens.scope || '').split(' ').filter(Boolean))
+    const missing = scopes.filter(s => !granted.has(s))
+    const expiry = tokens.expiry_date ? new Date(tokens.expiry_date) : null
+    const isExpired = expiry && expiry.getTime() < Date.now()
+
+    if (missing.length > 0 || isExpired) {
+      tokens = null
+    }
+  }
+
   if (!tokens) {
-    throw new Error(
-      'No Google credentials configured. Either:\n' +
-        '  - Run `mcp auth login` to authenticate as yourself (recommended for local use), or\n' +
-        '  - Set GOOGLE_APPLICATION_CREDENTIALS to a service-account JSON key file (Cloud Run / hosted), or\n' +
-        '  - Send a Google-issued bearer token in the Authorization header (HTTP transport).',
-    )
+    const isStdio = process.env.GCP_STDIO !== 'false' && !process.env.PORT
+    if (isStdio) {
+      const { startToolAuth, canLaunchBrowser } = await import('./credential/auth_login.js')
+      if (canLaunchBrowser()) {
+        if (apiOptions.onStatusUpdate) {
+          apiOptions.onStatusUpdate('Authentication required. Opening browser for Google OAuth consent...')
+        }
+        const result = await startToolAuth({ scopes })
+        if (result.status === 'completed') {
+          if (apiOptions.onStatusUpdate) {
+            apiOptions.onStatusUpdate('Authorization successful. Tokens cached.')
+          }
+          tokens = await cache.readEnforcingMode()
+        } else {
+          throw new Error('Sign-in did not complete inline. Run `mcp auth login` or `cep_auth`.')
+        }
+      } else {
+        throw new Error('Headless environment detected. Please run `cep_auth` tool or `mcp auth login`.')
+      }
+    } else {
+      throw new Error(
+        'No Google credentials configured. Either:\n' +
+          '  - Run `mcp auth login` to authenticate as yourself (recommended for local use), or\n' +
+          '  - Set GOOGLE_APPLICATION_CREDENTIALS to a service-account JSON key file (Cloud Run / hosted), or\n' +
+          '  - Send a Google-issued bearer token in the Authorization header (HTTP transport).',
+      )
+    }
   }
   const client = new OAuth2Client()
   client.setCredentials(tokens)
--- a/lib/util/auth.js
+++ b/lib/util/auth.js
@@ -29,6 +29,7 @@ limitations under the License.
 import fs from 'node:fs/promises'
 import { JWT, OAuth2Client } from 'google-auth-library'
 import { TokenCache } from './credential/token_cache.js'
+import { isStdioMode } from './gcp.js'
 import { SCOPES } from '../constants.js'
 
 /**
@@ -129,14 +130,13 @@ export async function getAuthClient(scopes, authToken, apiOptions = {}) {
   }
 
   if (!tokens) {
-    const isStdio = process.env.GCP_STDIO !== 'false' && !process.env.PORT
-    if (isStdio) {
+    if (isStdioMode()) {
       const { startToolAuth, canLaunchBrowser } = await import('./credential/auth_login.js')
       if (canLaunchBrowser()) {
         if (apiOptions.onStatusUpdate) {
           apiOptions.onStatusUpdate('Authentication required. Opening browser for Google OAuth consent...')
         }
-        const result = await startToolAuth({ scopes: Object.values(SCOPES) })
+        const result = await startToolAuth({ scopes: scopes.length > 0 ? scopes : Object.values(SCOPES) })
         if (result.status === 'completed') {
           if (apiOptions.onStatusUpdate) {
             apiOptions.onStatusUpdate('Authorization successful. Tokens cached.')
--- a/lib/util/credential/auth_login.js
+++ b/lib/util/credential/auth_login.js
@@ -56,9 +56,14 @@ let pendingAuth = null
  * @param {object} [opts] Optional overrides for tests.
  * @param {number} [opts.now] The "now" timestamp in ms; defaults to Date.now().
  * @param {string} [opts.cachePath] The cache file path; defaults to TokenCache.defaultPath().
+ * @param {string[]} [opts.scopes] Optional scopes to check for coverage.
  * @returns {Promise<TokenValidity>} Whether the cache holds a usable token, and why not when not.
  */
-export async function isTokenLocallyValid({ now = Date.now(), cachePath = TokenCache.defaultPath() } = {}) {
+export async function isTokenLocallyValid({
+  now = Date.now(),
+  cachePath = TokenCache.defaultPath(),
+  scopes = [],
+} = {}) {
   const cache = new TokenCache(cachePath)
   const tokens = await cache.read()
   if (!tokens) {
@@ -67,13 +72,25 @@ export async function isTokenLocallyValid({ now = Date.now(), cachePath = TokenC
   if (!tokens.access_token) {
     return { ok: false, reason: 'malformed' }
   }
-  if (!tokens.expiry_date) {
-    return { ok: true, expiresAt: null }
+
+  // 1. Check expiration first (most common and unambiguous failure)
+  if (tokens.expiry_date) {
+    const expiresAt = new Date(tokens.expiry_date)
+    if (expiresAt.getTime() < now) {
+      return { ok: false, reason: 'expired', expiresAt }
+    }
   }
-  const expiresAt = new Date(tokens.expiry_date)
-  if (expiresAt.getTime() < now) {
-    return { ok: false, reason: 'expired', expiresAt }
+
+  // 2. Check for scope coverage
+  if (scopes.length > 0) {
+    const granted = new Set((tokens.scope || '').split(' ').filter(Boolean))
+    const missing = scopes.filter(s => !granted.has(s))
+    if (missing.length > 0) {
+      return { ok: false, reason: 'insufficient' }
+    }
   }
+
+  const expiresAt = tokens.expiry_date ? new Date(tokens.expiry_date) : null
   return { ok: true, expiresAt }
 }
 
@@ -189,6 +206,7 @@ export async function startToolAuth({
   const callbackPromise = server.waitForCode().then(r => ({ kind: 'callback', ...r }))
   let timeoutHandle
   const timeoutPromise = new Promise(resolve => {
+    // No browser opened means no callback can arrive, so we skip the wait to avoid process hang.
     const effectiveWaitMs = browserAttempted ? awaitCallbackMs : 0
     timeoutHandle = setTimeout(() => resolve({ kind: 'timeout' }), effectiveWaitMs)
   })
--- a/lib/util/credential/auth_login.js
+++ b/lib/util/credential/auth_login.js
@@ -189,7 +189,8 @@ export async function startToolAuth({
   const callbackPromise = server.waitForCode().then(r => ({ kind: 'callback', ...r }))
   let timeoutHandle
   const timeoutPromise = new Promise(resolve => {
-    timeoutHandle = setTimeout(() => resolve({ kind: 'timeout' }), awaitCallbackMs)
+    const effectiveWaitMs = browserAttempted ? awaitCallbackMs : 0
+    timeoutHandle = setTimeout(() => resolve({ kind: 'timeout' }), effectiveWaitMs)
   })
   const winner = await Promise.race([callbackPromise, timeoutPromise])
   clearTimeout(timeoutHandle)
--- a/lib/util/credential/token_cache.js
+++ b/lib/util/credential/token_cache.js
@@ -75,6 +75,20 @@ export class TokenCache {
   }
 
   /**
+   * Deletes the cache file if it exists.
+   * @returns {Promise<void>} Resolves when the file has been deleted.
+   */
+  async clear() {
+    try {
+      await fs.unlink(this._path)
+    } catch (err) {
+      if (err.code !== 'ENOENT') {
+        throw err
+      }
+    }
+  }
+
+  /**
    * Returns true when the cache file's mode is exactly 0600 (no group/other access).
    * @returns {Promise<boolean>} True if the file mode is 0600, false otherwise.
    */
--- a/lib/util/gcp.js
+++ b/lib/util/gcp.js
@@ -28,6 +28,22 @@ import { logger } from './logger.js'
 import axios from 'axios'
 
 /**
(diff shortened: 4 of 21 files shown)


lib/util/auth.js
```diff
@@ -113,14 +114,36 @@ export async function getAuthClient(scopes, authToken) {
...
+      const { oauthFlowCredential } = await import('./credential/oauth_flow.js')
+      const cred = oauthFlowCredential({ requiredScopes: scopes })
+      tokens = await cred.runLoginFlow({ onStatusUpdate: apiOptions.onStatusUpdate })
+    } else {
+      throw new Error(
+        'No Google credentials configured. Either:\n' +
+          '  - Run `mcp auth login` to authenticate as yourself (recommended for local use), or\n' +
+          '  - Set GOOGLE_APPLICATION_CREDENTIALS to a service-account JSON key file (Cloud Run / hosted), or\n' +
+          '  - Send a Google-issued bearer token in the Authorization header (HTTP transport).',
+      )
```
timfee, line 157: This is the core mechanic that conflicts with merged #211. After the rebase, `guardedToolCall` (in `tools/utils/wrapper.js`) does its own pre-flight: reads the cache, checks expiry, and short-circuits with a structured `{ status: 'authRequired', nextAction: 'invoke-cep_auth' }` response *before* the handler runs. So this inline `runLoginFlow` branch never executes — by the time `getAuthClient` is called, either the cache is fresh or the wrapper already returned.

If you want to keep the single-turn UX, the discussion is whether it should replace 211's pre-flight model or coexist. A few concerns with the inline-blocking approach as-is:
- A tool call can hang for the consent timeout, which exceeds many MCP client deadlines. 211 caps the wait at 90s (`CEP_AUTH_TIMEOUT_MS`).
- Doesn't cover headless: 211 falls back to a paste-back URL (`status: awaiting`, `nextAction: paste-redirect-url`). Here the call just hangs until the loopback server times out.
- Concurrent expired-auth tool calls race on `getAuthClient`; would each open its own browser tab.

Simplest path: drop this branch on rebase and rely on the merged pre-flight. Keep the `onStatusUpdate` plumbing — it's complementary.

lib/util/gcp.js
```diff
@@ -50,6 +50,9 @@ async function fetchMetadata(path) {
  * @returns {Promise<{project: string, region: string}|null>} An object containing project and region, or null if not available
  */
 export async function checkGCP() {
+  if (process.env.GCP_STDIO === 'true' || process.env.CEP_BACKEND === 'fake') {
+    return null
+  }
```
timfee: This short-circuit is great and orthogonal to the auth changes — fixes the 3s metadata hang on stdio / fake-backend test runs. Worth landing even if the rest of the PR gets restructured.

lib/util/credential/token_cache.js
```diff
@@ -74,6 +74,20 @@ export class TokenCache {
...
+  async clear() {
+    try {
+      await fs.unlink(this._path)
+    } catch (err) {
+      if (err.code !== 'ENOENT') {
+        throw err
+      }
+    }
+  }
+
```
timfee, line 90: `clear()` is a clean addition and pairs naturally with the `auth_clear` tool. Keep.

timfee (commented):
Heads-up: #211 (`feat/auth-preflight-tool`) merged into main earlier today (sha 9b75909a). It implements an explicit pre-flight + `cep_auth` model that conflicts with the inline auto-auth model here — see the comment on `lib/util/auth.js`. Worth rebasing onto main and adapting this PR to add what 211 doesn't have, rather than re-deciding the auth model.

Where this PR has clear additive value over 211:
- `auth_status` and `auth_clear` tools (211 only adds `cep_auth`)
- `TokenCache.clear()` helper
- `checkGCP()` short-circuit for stdio / fake backend (a real bug fix, independent of auth)
- `onStatusUpdate` plumbing for `server.sendLoggingMessage` (nice UX even with the pre-flight model)

Where this PR conflicts with 211:
- The inline `getAuthClient()` that calls `runLoginFlow` mid-handler vs 211's pre-flight in `guardedToolCall`. Pick one — running both means whichever fires first wins and the other is dead code.

Flagged a few hygiene items inline.

timfee (commented):
Thanks for the cleanup pass and the rebase — branch is sitting on top of #211 cleanly now.

One thing left from the model-conflict comment on `lib/util/auth.js`: the inline path in `getAuthClient()` runs *after* #211's pre-flight short-circuits in `guardedToolCall`, so for the missing/expired-token case the inline branch is unreachable. The scope-gap case it covers is genuinely new (#211's pre-flight doesn't check scope coverage), but the action — block the tool call until consent completes — has the UX issues from the earlier review (no headless fallback, can exceed MCP client deadlines, concurrent calls race). Worth deciding whether to: drop the inline branch and rely on a follow-up to extend the pre-flight to scope-gap detection, or keep the inline branch but bound it (cap timeout, fall back to a paste-back response on headless).

On the zod side: #223 already runs the v4 codemod across the codebase, including #211's `tools/definitions/auth.js`. When #223 merges, your `auth_status` / `auth_clear` additions will still use the deprecated `.passthrough()` form on a v4 main. Happy to run the codemod on your two new additions and push the commit to your branch once #223 lands — say the word.

Follow-ups for #211 are tracked separately in #229.

tools/utils/wrapper.js
```diff
@@ -39,26 +39,36 @@ function buildAuthRequiredResponse({ reason, expiresAt }) {
...
+  }
+
+  return {
+    content: [
+      { type: 'text', text },
+      // We include the structured data as a "hidden" code block in the content.
+      // This allows the agent (Gemini) to parse the signal if its internal logic needs it,
+      // while bypassing the rigid MCP Client SDK outputSchema validator which would
+      // fail if we returned it in structuredContent (since it doesn't match the tool's schema).
+      { type: 'text', text: '```json\n' + JSON.stringify(authRequiredData, null, 2) + '\n```' },
```
timfee: The new `content[]` block puts the `authRequired` payload into a Markdown code fence to avoid the `outputSchema` validator that crashed on the original `structuredContent`. The crash diagnosis is correct, but typed MCP clients can't parse Markdown to recover `authRequired`.

Three fixes keep the structured signal intact: declare the tool's `outputSchema` as a union that admits `authRequired`; skip schema enforcement when `isError: true`; or move `authRequired` to a top-level key the SDK doesn't validate.

timfee (commented):
Commit 7ad4fcb fixes a real SDK crash but trades the typed `authRequired` contract for a JSON code fence inside `content[]`. That's the one thing I'd want resolved before merge.

The validator threw because `structuredContent` didn't match the tool's declared `outputSchema`. The diagnosis is correct. The fix as-implemented forces typed MCP clients to parse Markdown to find `authRequired`, which they won't do.

Three fixes keep the structured signal intact: declare each guarded tool's `outputSchema` as a union that admits `authRequired`; skip schema enforcement when `isError: true`; or move `authRequired` to a top-level key the SDK doesn't validate.

The two unit tests updated to `assert.ok(!result.structuredContent)` now only confirm the old shape is absent. Whichever fix we pick above, those tests should assert the agent can still recover `reason`, `nextAction`, and `expiresAt`.

One smaller item is worth fixing before ship. The stdio detection in `getAuthClient` is `process.env.GCP_STDIO !== 'false' && !process.env.PORT`. A cloud-run deploy that forgets to set `GCP_STDIO=false` evaluates true, lands in the inline-OAuth path, and surfaces "Headless environment detected" instead of the previous "no credentials configured" message. Lifting the condition into a named `isStdioMode()` plus a test for the cloud-run env combination would catch the regression.

Three follow-ups are deliberately out of scope and worth tracking: consume the `AUTH_STATUS` and `AUTH_NEXT_ACTION` constants from #230 once it merges; pick up the `cep_auth_status` / `cep_auth_clear` rename and tightened descriptions from #232 for the Workspace MCP namespace collision Shantanu hit; tighten the `cep_auth` description per #231.

The rest of the diff is fine. The scope-aware cache check is a real improvement; previously, under-scoped tokens failed at the API layer with a confusing 401. `TokenCache.clear`, the `checkGCP` short-circuit, and the `onStatusUpdate` plumbing all look right.

test/local/utils.test.js
```diff
@@ -341,9 +341,7 @@ describe('Tool Utils', () => {
         assert.strictEqual(result.isError, true)
         assert.match(result.content[0].text, /Sign-in is needed/)
         assert.match(result.content[0].text, /cep_auth/)
-        assert.strictEqual(result.structuredContent.authRequired.reason, 'missing')
-        assert.strictEqual(result.structuredContent.authRequired.nextAction, 'invoke-cep_auth')
-        assert.match(result.structuredContent.authRequired.docsUrl, /configuration\.md#authenticate-to-google-apis/)
+        assert.ok(!result.structuredContent)
```
timfee, line 176: Both this assertion and the one at line 370 now only confirm `structuredContent` is undefined. If we settle the `outputSchema` question by putting the `authRequired` payload somewhere the agent can still read (a union schema, a different top-level key), these tests pass regardless of whether the new path is wired correctly.

The replacement should exercise the agent-visible shape, asserting that `reason`, `nextAction`, and `expiresAt` are recoverable.

lib/util/auth.js
```diff
@@ -113,14 +115,47 @@ export async function getAuthClient(scopes, authToken) {
...
-        '  - Send a Google-issued bearer token in the Authorization header (HTTP transport).',
-    )
+    const isStdio = process.env.GCP_STDIO !== 'false' && !process.env.PORT
+    if (isStdio) {
+      const { startToolAuth, canLaunchBrowser } = await import('./credential/auth_login.js')
+      if (canLaunchBrowser()) {
+        if (apiOptions.onStatusUpdate) {
+          apiOptions.onStatusUpdate('Authentication required. Opening browser for Google OAuth consent...')
+        }
+        const result = await startToolAuth({ scopes: Object.values(SCOPES) })
```
timfee: `getAuthClient` receives `scopes` as its first argument, but this call to `startToolAuth` ignores it and substitutes `Object.values(SCOPES)`. The cache stores the full SCOPES superset today, so the behavior happens to be right, but the next reader will assume `scopes` has effect on every path.

Either pass `scopes` through, or document that SCOPES is the cache contract on this path.

lib/util/auth.js
```diff
@@ -113,14 +115,47 @@ export async function getAuthClient(scopes, authToken) {
...
+  }
+
   if (!tokens) {
-    throw new Error(
-      'No Google credentials configured. Either:\n' +
-        '  - Run `mcp auth login` to authenticate as yourself (recommended for local use), or\n' +
-        '  - Set GOOGLE_APPLICATION_CREDENTIALS to a service-account JSON key file (Cloud Run / hosted), or\n' +
-        '  - Send a Google-issued bearer token in the Authorization header (HTTP transport).',
-    )
+    const isStdio = process.env.GCP_STDIO !== 'false' && !process.env.PORT
```
timfee: This condition evaluates true for any environment missing both `GCP_STDIO=false` and `PORT`, including a cloud-run deploy where someone forgot to set `GCP_STDIO=false`. That deploy lands in the inline-OAuth path and throws "Headless environment detected" instead of the previous "no credentials configured" error.

Lift the condition into a named `isStdioMode()` helper and add a test covering the cloud-run env combination (no `GCP_STDIO`, no `PORT`, no `GOOGLE_APPLICATION_CREDENTIALS`).

lib/util/credential/auth_login.js
```diff
@@ -189,7 +189,8 @@ export async function startToolAuth({
   const callbackPromise = server.waitForCode().then(r => ({ kind: 'callback', ...r }))
   let timeoutHandle
   const timeoutPromise = new Promise(resolve => {
-    timeoutHandle = setTimeout(() => resolve({ kind: 'timeout' }), awaitCallbackMs)
+    const effectiveWaitMs = browserAttempted ? awaitCallbackMs : 0
```
timfee, line 210: Setting the callback wait to zero milliseconds when `browserAttempted` is false made me stop and check whether the loopback timeout was disabled accidentally. The intent looks correct. No browser opened means no callback can arrive, so the wait would only delay the paste-back path.

A one-line comment to that effect would save the next reader the round trip.

tools/utils/wrapper.js
```diff
@@ -166,11 +177,24 @@ export function guardedToolCall(
...
+        const onStatusUpdate = msg => {
+          try {
+            if (typeof server?.sendLoggingMessage === 'function') {
+              server.sendLoggingMessage({ level: 'info', data: msg }).catch(() => {})
+            }
+          } catch {
+            // ignore
+          }
+        }
+        options.apiOptions = { ...(options.apiOptions || {}), onStatusUpdate }
```
timfee: This writes onto the same `options.apiOptions` object the caller passed in, so every subsequent tool invocation sharing that reference inherits the `onStatusUpdate` added by the first call. `registerTools` passes a single options instance and the callback is idempotent, so it's safe today.

A short comment explaining the intent, or a local destructure, would prevent a future caller from being surprised.

TusharKanekiDey (approved):
Can you add the documentation you referred here or in the bug for more details.

lib/util/auth.js
dylan-klein: I have implemented isStdioMode in lib/util/gcp.js as suggested. This more robustly distinguishes between local usage and Cloud Run, ensuring we do not accidentally try to pop a browser in headless or containerized modes. I have also added a dedicated test covering the Cloud Run environment combinations to prevent regressions.

tools/utils/wrapper.js
dylan-klein: Regarding the choice of a text-only error for the Auth Required signal instead of using z.union() or a dedicated schema field: I explicitly moved away from those after hitting some technical blockers in the current MCP SDKs. The server-side SDK currently crashes with a TypeError during initialization if it encounters a z.union() or non-object schema. Furthermore, the Gemini CLI middleware validates the structuredContent of a response against the tools success schema even when isError: true is set. By adopting a Strict Success, Unstructured Error model, we keep our success schemas clean and focused purely on data results without polluting every contract with failure-state metadata. Since Gemini can already autonomously parse the remediation instructions (like the command to run mcp auth login) from the error text, this approach bypasses the SDK bugs while maintaining full agent functionality.

test/local/utils.test.js
dylan-klein: To ensure the build stays green and the auth logic is fully verified, I have hardened the test runners to propagate a fully scoped synthetic environment so the suite remains decoupled from the developers local token cache. I have also updated the specific unit tests to correctly simulate a headless environment so they take the expected signaling path. All 447 unit tests and 31 integration tests are now passing with these refinements.

dylan-klein:
I have finalized the implementation with the requested technical and architectural refinements. 

Key changes in the latest push:
1. Implemented isStdioMode in lib/util/gcp.js to robustly handle environment detection across local CLI and Cloud Run.
2. Restored the strict pre-flight security check in the tool wrapper. All environments, including fake modes, now enforce valid authentication.
3. Hardened the test suite by decoupling unit tests via mock Inbound Bearer tokens and updating test runners to propagate a fully scoped synthetic environment. 
4. Resolved the Zod validation issue by adopting a Strict Success, Unstructured Error model. This maintains strict data contracts for success paths while bypassing SDK bugs for authentication remediation.

dylan-klein:
Regarding the documentation reference: I primarily referred to the implementation in the Gemini CLI Workspace Extension (`google3/third_party/javascript/node_modules/google_gemini_cli/bundle/builtin/google-workspace`) as the blueprint for the inline authentication and loopback server logic. I also consulted the Model Context Protocol SDK documentation for the transport-level constraints.

lib/util/auth.js
dylan-klein: I have updated getAuthClient to pass the requested scopes through to startToolAuth, falling back to the full SCOPES set only if the input is empty. This ensures the granular scope logic is respected on this path as well.

lib/util/credential/auth_login.js
dylan-klein: Added the suggested comment explaining that the wait is skipped when no browser was opened because no loopback callback can arrive.

tools/utils/wrapper.js
dylan-klein: Good catch. I have updated the tool wrapper to use a local apiOptions variable instead of mutating the shared options object, preventing unintended inheritance across different tool calls.

timfee:
Thanks for the turnaround. Moving the stdio check into `isStdioMode` covers the concern. The per-tool `scopes` argument through `isTokenLocallyValid` is a direction I hadn't considered, and checking scope coverage in the wrapper catches under-scoped caches earlier than the inline path would.

On the unstructured-error choice, the two SDK bugs you named close off the union and skip-on-isError options. One thing that would help anyone returning to this code: a comment on `buildAuthRequiredResponse` naming both bugs (with upstream issue links if any are filed), so when one of them gets fixed it's obvious we can revisit.

Two small items from the first pass that I think are still open, both code-comment-only: the `options.apiOptions` in-place mutation in `wrapper.js`, and the `effectiveWaitMs = browserAttempted ? awaitCallbackMs : 0` line in `auth_login.js`. A short comment or a local destructure on the first would help the next reader.

Three follow-ups for after this merges so they don't slip: consuming `AUTH_STATUS` and `AUTH_NEXT_ACTION` from #230, picking up the `cep_auth_status` / `cep_auth_clear` rename and tightened descriptions from #232 for the Workspace MCP namespace collision Shantanu hit, and the `cep_auth` description tightening in #231.

lib/util/auth.js
timfee: Thanks, that covers what I was worried about.

tools/utils/wrapper.js
timfee: Got it on the SDK constraints; that explains why the union and skip-on-isError options weren't available. A comment on `buildAuthRequiredResponse` naming the two bugs (the `z.union()` init crash and the middleware that validates `structuredContent` against the success schema even when `isError: true`) would let anyone returning to this code know the choice was forced by external bugs rather than a preference. Links to upstream issues would let us notice when either bug is fixed.

test/local/utils.test.js
timfee: One thought on the `SSH_CONNECTION=true` line in the test: it depends on whatever signals `canLaunchBrowser` currently checks for. If that helper ever picks up another signal, the test silently stops covering this path. Mocking `canLaunchBrowser` directly would avoid the coupling. Small thing, not asking you to redo it.

tools/utils/wrapper.js
dylan-klein: I have added a detailed comment to `buildAuthRequiredResponse` naming both the Server SDK crash (serialization of unions) and the Client SDK validation rigidity (checking structuredContent on errors) as the reasons for this approach. This should make the rationale clear for future maintainers. I have also verified that the `apiOptions` isolation and the `auth_login.js` comment are correctly implemented in the latest push. Regarding the follow-ups for constants, renaming, and descriptions: acknowledged, I will leave those for the subsequent PRs so we can get this core logic landed.

dylan-klein:
I've added the requested documentation comment to buildAuthRequiredResponse in tools/utils/wrapper.js naming the specific SDK bugs. I've also verified that the apiOptions isolation and the auth_login.js timeout comment are correctly implemented in the latest push. Thanks for the thorough review!

PR #232, opened 2026-05-16
fix(auth): cep_ prefix on auth_status/auth_clear; styled loopback pages

timfee:
Targeted at Dylan's #228 branch so it lands together with the proactive inline-auth work.

## Summary
- Renames `auth_status` → `cep_auth_status` and `auth_clear` → `cep_auth_clear` so a client hosting both the CEP and Workspace MCP servers does not collide on generic `auth_*` tool names. Tool descriptions now name CEP MCP explicitly.
- Replaces the single plain-text loopback response with a small Material-styled success page and a matching error page. The error page is rendered when the consent redirect carries an `error` query parameter and includes the (escaped) OAuth error reason.
- Each rendered page is under 1.2 KB with inline CSS and no external requests.

## Test plan
- [x] `npm test` — all unit and integration suites green locally
- [ ] Sanity check the success page in a real browser after running the loopback flow end-to-end
- [ ] Sanity check the error page by clicking "Cancel" at the consent screen

### DIFF
--- a/lib/util/credential/loopback_server.js
+++ b/lib/util/credential/loopback_server.js
@@ -38,13 +38,12 @@ export async function startLoopbackServer() {
 
   const server = http.createServer((req, res) => {
     const url = new URL(req.url, 'http://127.0.0.1')
-    res.writeHead(200, { 'Content-Type': 'text/html' })
-    res.end('<html><body><p>You may close this window.</p></body></html>')
-    resolveCode({
-      code: url.searchParams.get('code') ?? undefined,
-      state: url.searchParams.get('state') ?? undefined,
-      error: url.searchParams.get('error') ?? undefined,
-    })
+    const code = url.searchParams.get('code') ?? undefined
+    const state = url.searchParams.get('state') ?? undefined
+    const error = url.searchParams.get('error') ?? undefined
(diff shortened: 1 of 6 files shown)


dylan-klein (approved):
Approved. I really like the new styled login success screen—it's a great UX improvement over the plain text response.

lib/util/credential/loopback_server.js
```diff
@@ -63,3 +62,70 @@ export async function startLoopbackServer() {
...
+ */
+function renderSuccessPage() {
+  return (
+    '<!doctype html><html lang="en"><head><meta charset="utf-8">' +
+    '<meta name="viewport" content="width=device-width,initial-scale=1">' +
+    '<title>Signed in to CEP MCP</title>' +
+    `<style>${PAGE_STYLE}</style></head><body>` +
+    '<div class="card"><div class="icon ok">✓</div>' +
+    '<h1>Signed in</h1>' +
+    '<p>Chrome Enterprise Premium MCP has your credentials. You can close this window and return to your terminal.</p>' +
```
dylan-klein, line 94: Optional: Consider changing "return to your terminal" to "return to the agent" or "return to your chat" to better reflect that this is typically used via an AI agent.

dylan-klein (approved):

PR #234, opened 2026-05-18
fix(ci): resolve zod v4 cache mismatch in daily evaluations

TusharKanekiDey:
Resolves GHA daily evaluations failure by upgrading Node.js runner version to 24 and adding a diagnostic Zod version check step.

Bug: b/514395431
Bug: b/493244463

### DIFF
--- a/.github/workflows/agent-evals.yml
+++ b/.github/workflows/agent-evals.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 24
           cache: npm
       - run: npm ci
 


package-lock.json
```diff
@@ -4467,30 +4467,6 @@
         "node": ">=0.6"
       }
     },
```
TusharKanekiDey, line 4469: So package-lock.json changes is required ?

```
No, the package-lock.json change is not required to fix this issue.

Why?
Zod v4 was already correctly locked: Before we made any changes, package-lock.json on the main branch already had Zod locked at the correct version (4.4.3).
The issue was purely environment/runner-specific: The GHA evaluations workflow was failing because the runner (running on Node 20) restored/resolved stale dependencies from cache, locking the runner's local installation to Zod v3.
What actually fixes the issue: Upgrading the workflow runner to Node 24 in .github/workflows/agent-evals.yml resolves the package resolution/cache mismatch, allowing npm ci to successfully install Zod v4.
Why is package-lock.json in the PR then?
When we ran npm install locally in our workspace environment, NPM pruned a stale/unused optional dev dependency (ts-declaration-location). This is a minor mechanical cleanup byproduct and is not required for the fix.

If you prefer a cleaner commit/PR with only the required .github/workflows/agent-evals.yml change, we can easily revert the package-lock.json change and update the PR!

```

dylan-klein (approved):

PR #235, opened 2026-05-18
chore: sync gemini-extension.json version range during formatting

TusharKanekiDey:
Automatically sync the package version range in gemini-extension.json args array during the npm run format script execution.

Bug: b/514415277


timfee (commented):
Release please **should** in theory keep this up to date - might be worth fixing there as a stable long term fix so `format` doesn't produce unintentional side effects.

timfee (approved):

timfee:
Should be synched via .release-please-manifest.json

package.json
```diff
@@ -29,7 +29,7 @@
     "eval:knowledge": "CEP_BACKEND=fake node test/evals/run.js --category knowledge",
     "eval:licensing": "CEP_BACKEND=fake node test/evals/run.js --category licensing",
     "eval:prompt": "CEP_BACKEND=fake node test/evals/run.js --category prompt",
-    "format": "prettier --write .",
+    "format": "node -e \"const p=JSON.parse(require('fs').readFileSync('package.json','utf8')); const e=JSON.parse(require('fs').readFileSync('gemini-extension.json','utf8')); e.mcpServers['chrome-enterprise-premium'].args[1]='@google/chrome-enterprise-premium-mcp@^'+p.version; require('fs').writeFileSync('gemini-extension.json',JSON.stringify(e,null,2)+'\\n','utf8')\" && prettier --write .",
```
dylan-klein, line 32: If the `npx` argument order changes in the future, will hardcoding `args[1]` here correctly identify the version string? Would moving this to a dedicated script file allow for more robust prefix-based matching and better maintainability than an inline one-liner?

dylan-klein (commented):
Since `npm run format` is frequently used as a pre-commit hook, will attaching this sync logic cause unrelated version bumps to appear in developer PRs whenever the main version increases? Should this instead be part of the release process to avoid "dirty" commits in daily development?

timfee:
Yeah I do worry about a side effect that isn't intended and would prefer we do via release-please more properly if possible

PR #240, opened 2026-05-19
fix(auth): pad authUrl line and tell agent to restate it in chat

### LINKED ISSUE #238: OAuth UX: long auth URL clips against Gemini CLI tool-output box border on copy/paste
## Problem

The `cep_auth` tool returns its consent URL inside the `content: [{type: 'text'}]` block of its tool response. Gemini CLI renders MCP tool output inside a Unicode-bordered box. When the URL ends within ~1 character of the right border, double-click selection or drag-select catches the border glyph and breaks the URL. The agent's restatement in normal chat sits outside the box and selects cleanly — which is why users who copy from the chat version succeed and users who copy from the tool-output box don't.

## Emission sites

- **`tools/definitions/auth.js:206`** — `awaitingResponse()` pushes `result.authUrl` as a bare line inside the `content` text block. **This is the box-render site.**
- `lib/util/credential/oauth_flow.js:272` — `process.stderr.write` of the consent URL during `mcp auth login`. Raw terminal, not Gemini CLI's box. **No change.**
- `oauth_flow.js:270` — `onStatusUpdate(consentUrl)` surfaces via MCP progress notifications. Lower priority.

Note: the URL is already exposed in `structuredContent.authUrl` (line 216), but Gemini CLI does not currently render structured content as a copyable link.

## Proposed approach

Single fix in `tools/definitions/auth.js`:

1. Tool description (`cep_auth`, ~line 62-64): append a sentence instructing the agent to restate the `authUrl` in its chat reply on its own line, outside any tool-output box, because long URLs can clip during selection inside boxed tool output.
2. `AGENT_HINT` constant (~line 35-39): mirror the same restate-in-chat instruction.
3. `awaitingResponse()` (`lines.push(result.authUrl)` at ~line 206): pad with two leading/trailing spaces. The padding gives selection a buffer from the box border.

Skip OSC 8 hyperlink escapes (Gemini CLI's web UI ignores them) and skip emitting only via `structuredContent` (that's a client-side ask).

## Acceptance criteria

- URL line in tool output has whitespace padding from box borders.
- Tool description and `AGENT_HINT` both tell the agent to restate the URL in chat.
- Tests in `test/local/auth-tool.test.js` (or wherever the description / `awaitingResponse` is asserted) cover the new format.

## Scope

Small. One file, ~6 lines + test updates.

timfee:
The `cep_auth` tool emits the consent URL inside an MCP text block, which Gemini CLI renders in a Unicode-bordered box where long URLs clip against the right border on selection. This pads the URL line with two leading and two trailing spaces and adds a restate-in-chat instruction to both the tool description and `AGENT_HINT`, so the agent mirrors the URL outside the box where selection is clean.

Closes #238

### DIFF
--- a/test/local/auth-tool.test.js
+++ b/test/local/auth-tool.test.js
@@ -82,8 +82,25 @@ describe('cep_auth Tool', () => {
     assert.strictEqual(result.structuredContent.nextAction, 'paste-redirect-url')
     assert.strictEqual(result.structuredContent.authUrl, 'https://accounts.google.com/o/oauth2/v2/auth?state=ABC')
     assert.ok(result.structuredContent.agentHint?.length > 0)
+    assert.match(result.structuredContent.agentHint, /restate.*chat/i)
+    assert.match(result.structuredContent.agentHint, /outside.*tool-output box/i)
     assert.match(result.content[0].text, /Open this URL/)
     assert.match(result.content[0].text, /accounts\.google\.com/)
+    assert.ok(
+      result.content[0].text.split('\n').includes('  https://accounts.google.com/o/oauth2/v2/auth?state=ABC  '),
+      'authUrl line should be padded with two leading and two trailing spaces',
+    )
+  })
+
+  test('When cep_auth is registered, then its description tells the agent to restate the authUrl in chat outside the tool-output box', async () => {
+    const startToolAuth = mock.fn()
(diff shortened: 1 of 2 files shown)


timfee:
Replacing with OSC 8 + own-line treatment per survey of mature CLI patterns. See #238 follow-up PR.

PR #241, opened 2026-05-19
feat(auth): boxed consent URL and Chrome activation on macOS

### LINKED ISSUE #236: OAuth UX: browser launch lacks focus signal; users miss the sign-in tab
## Problem

Tester feedback (Tushar): when `mcp auth login` or the `cep_auth` tool launches the system browser, the consent tab opens inside an existing browser window which may not have focus or may be behind another Space/desktop. The terminal prints the consent URL on a normal-weight line, which is easy to scroll past in a long shell. Users miss the sign-in tab entirely and the flow stalls.

## Current behavior

`defaultOpenBrowser` in `lib/util/credential/oauth_flow.js:34` spawns `open` / `xdg-open` / `cmd start` directly. No activation hint, no failure detection beyond exit code, no terminal "now sign in" beacon.

## Proposed approach

Roll our own — do not pull the `open` npm package. `open@v10` requires Node ≥20, this repo's `engines.node` is `>=18`. The version of `open` that supports Node 18 has no useful focus knob and adds 6+ transitive deps.

In `lib/util/credential/oauth_flow.js`:

1. On macOS: when `$BROWSER` is unset and Chrome is installed, prefer `open -a "Google Chrome" <url>`. Activating a named app reliably brings it forward. Cache the Chrome-detection result per process. Fall back to plain `open <url>`.
2. On Windows: keep `cmd /c start "" <url>`.
3. On Linux: keep `xdg-open`. No `wmctrl` dep.
4. Detect remote/container before launch (`SSH_CONNECTION`, `SSH_TTY`, `/.dockerenv`, `/proc/1/cgroup` contains `docker|kubepods`). When remote, return `false` from `openBrowser` without spawning so the existing paste-the-URL fallback in `auth_login.js:200` and `oauth_flow.js:273` kicks in.
5. Replace the current pre-launch URL print with a bold, boxed, ANSI-coloured block (Unicode box drawing, bright cyan), gated on `process.stderr.isTTY`. Keep the plain URL line for non-TTY callers (logs, MCP transports).

## Acceptance criteria

- Browser launch on macOS with Chrome installed and `$BROWSER` unset activates Chrome.
- Headless/SSH session: no browser spawn attempted; the user sees the paste-URL prompt.
- TTY shell: the URL block is visually distinct (boxed/coloured).
- Non-TTY (MCP transport): output unchanged.
- Existing tests in `test/local/credential-oauth-flow.test.js` and `test/local/auth_login.test.js` continue to pass.

## Scope

Small. ~60 lines in `oauth_flow.js`, no API changes, existing injection points cover testing.

timfee:
## What this changes

`defaultOpenBrowser` now calls the `open` npm package instead of spawning `open` / `xdg-open` / `cmd start` directly. The platform-specific activation calls added in an earlier iteration (osascript on macOS, PowerShell `AppActivate` on Windows, `wmctrl` on Linux, OSC 9 desktop notification) are gone.

## Why

Forcing the browser to the foreground is not reliably solvable from the terminal side. The OS controls activation. The previous activation stack worked in the lucky cases and silently failed everywhere else, which is the worst of both worlds — fragile code that gave a false sense of completeness.

A survey of widely-used CLIs that face the same problem confirmed the boring answer: open the URL, print a clear terminal cue, accept that the user may need to click to focus the tab. None of `gh`, `gcloud`, `vercel`, `firebase`, or `npm` try to force focus; all of them rely on a prominent URL line and a brief instruction.

## What stayed

The terminal cue is still useful, so the parts of the earlier iteration that don't depend on OS activation remain:

- `printConsentUrl` renders a bright-cyan Unicode box on TTY stderr.
- The URL inside the box is wrapped in an OSC 8 hyperlink escape, so terminals that support it (iTerm2, WezTerm, kitty, Windows Terminal, VS Code, Ghostty) render it as clickable.
- A single BEL plays on TTY launch as an audible cue.
- `canLaunchBrowser` still short-circuits the launch in headless environments (SSH, no `DISPLAY`, container-via-cgroup), so the paste-back path is reached without a failed launch attempt.

## Prior art

The `open` package is used by [Vite](https://github.com/vitejs/vite), [Vercel CLI](https://github.com/vercel/vercel), [Storybook](https://github.com/storybookjs/storybook), and many others for the same cross-platform browser-launch problem. It handles platform detection, `$BROWSER` env-var honoring, and graceful failure.

## Engine bump

`open@v10` requires Node `>=20`. `engines.node` moves from `>=18.0.0` to `>=20.0.0`. Node 18 went EOL in April 2025.

## Tests

Unit tests cover the launcher-injection point, the headless short-circuit, the BEL on TTY, and the OSC 8 wrap. `npm run test:unit` and `npm run lint` both pass.

Closes #236.

### DIFF
--- a/gemini-extension.json
+++ b/gemini-extension.json
@@ -6,10 +6,7 @@
   "mcpServers": {
     "chrome-enterprise-premium": {
       "command": "npx",
-      "args": [
-        "-y",
-        "@google/chrome-enterprise-premium-mcp@^1.4.0"
-      ]
+      "args": ["-y", "@google/chrome-enterprise-premium-mcp@^1.4.0"]
     }
   }
 }
--- a/lib/util/credential/auth_login.js
+++ b/lib/util/credential/auth_login.js
@@ -31,6 +31,7 @@ limitations under the License.
    after-await pattern that triggers this rule is the intended behavior. */
 
 import { randomBytes } from 'node:crypto'
+import fsSync from 'node:fs'
 import { OAuth2Client } from 'google-auth-library'
 import { TokenCache } from './token_cache.js'
 import { startLoopbackServer } from './loopback_server.js'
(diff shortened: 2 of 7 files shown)


dylan-klein:
PR title mismatch: the implementation covers container detection and Shell UX (OSC 8 links, boxes) rather than Device Trust tools.

dylan-klein (approved):

PR #247, opened 2026-05-19
chore: format gemini-extension.json

### LINKED ISSUE #248: gemini-extension.json drifts out of prettier format because tooling doesn't reformat it
## Problem

The Kokoro presubmit check runs \`prettier --check .\` across the whole repo. \`gemini-extension.json\` periodically drifts out of prettier format and trips the check on \`main\`, blocking every PR. \`lint-staged\` (pre-commit hook) only formats staged files, so a contributor whose change doesn't touch \`gemini-extension.json\` can't fix the drift incidentally.

## Sources of drift

1. **release-please** — \`release-please-config.json\` lists \`gemini-extension.json\` as an extra-files target with \`jsonpath: \$.version\`. On every release, release-please rewrites the version string and commits without running prettier.
2. **Manual / scripted edits to the \`args\` version range** — PR #235 was titled "chore: sync gemini-extension.json version range during formatting", suggesting there's a (manual? scripted?) workflow that updates the \`@^x.y.z\` range inside the \`args\` array. Whatever produces that change also doesn't re-run prettier.

## Options

1. **Pre-commit hook runs \`prettier --check .\`** in addition to \`lint-staged\`. Catches whole-tree drift before commit. Slower commits (a few seconds), but enforces invariant.
2. **CI job that auto-formats and pushes** — a GitHub Action that runs \`prettier --write .\` on \`main\` and opens a PR (or commits directly) when drift is found. Heavier but recovers from drift automatically.
3. **Patch release-please** — add a post-commit \`prettier --write\` step (release-please supports custom postprocess hooks via \`extra-files\` config in recent versions, or via a separate workflow that re-runs after the release PR is created).
4. **Add a \`prettier --write\` step to whatever script syncs the \`args\` version range** if such a script exists. Otherwise document the manual flow.

## Acceptance criteria

- \`prettier --check .\` passes on \`main\` after every release-please run.
- \`prettier --check .\` passes on \`main\` after every \`args\` version-range sync.
- Kokoro presubmit no longer fails on the prettier check.

## Immediate workaround

PR opened to fix the current drift: \`chore: format gemini-extension.json\`.

timfee:
Unblocks the Kokoro presubmit check, which runs \`prettier --check .\` across the whole tree and currently fails on \`main\` because \`gemini-extension.json\`'s \`args\` array is split across three lines while prettier wants it inlined. Bringing the file back in line with prettier defaults.

\`lint-staged\` only formats staged files, so unrelated changes don't catch drift in files the contributor didn't touch — that's how this slipped through. A follow-up issue tracks the root cause: the file is rewritten by tooling (release-please version bumps; possibly an args-version sync script) without a subsequent prettier pass.

### DIFF
--- a/gemini-extension.json
+++ b/gemini-extension.json
@@ -6,10 +6,7 @@
   "mcpServers": {
     "chrome-enterprise-premium": {
       "command": "npx",
-      "args": [
-        "-y",
-        "@google/chrome-enterprise-premium-mcp@^1.4.0"
-      ]
+      "args": ["-y", "@google/chrome-enterprise-premium-mcp@^1.4.0"]
     }
   }
 }


dylan-klein (approved):
Approved. Correctly syncs the version in `gemini-extension.json` to 1.6.0 and cleans up formatting.

PR #254, opened 2026-05-19
fix(evals): seed synthetic valid token in evals runner for fake backend

TusharKanekiDey:
Bypasses the tool call pre-flight token cache check in the daily evaluations runner when running under the fake backend. Seeding a synthetic tokens.json file prevents all tool-using evaluations from stalling or failing due to missing cached OAuth credentials in the clean GitHub Actions runner environment.

### Verification Completed
* Ran local dry-run evaluations with the local credentials cache temporarily removed/moved. All **103 of 103 evaluations passed cleanly (100.0% success rate)**.
* This confirms that the synthetic `tokens.json` seeding is fully functional and correctly redirects the MCP tool wrapper pre-flight authentication check.
* Verified all lints, style checks, unit/integration/smoke tests passed successfully.

Bug: b/514746606
Bug: b/514395431
Bug: b/493244463

### DIFF
--- a/test/evals/run.js
+++ b/test/evals/run.js
@@ -51,6 +51,9 @@ import { parseArgs } from 'node:util'
 import path from 'node:path'
 import { fileURLToPath } from 'node:url'
 import fs from 'node:fs/promises'
+import fsSync from 'node:fs'
+import os from 'node:os'
+import { SCOPES } from '../../lib/constants.js'
 
 import { loadAllEvals, loadGlobalConfig } from './lib/loader.js'
 import { runChecks, checkForbidden, checkRequired } from './lib/assertions.js'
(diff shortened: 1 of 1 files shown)


timfee (approved):
LGTM. Same pattern as `test/run-unit.js:46-68` and `test/run-integration.js:50-70`; the evals runner was the missing one.

What I verified:
- `fsSync` disambiguation is necessary because `test/evals/run.js` already imports `fs from 'node:fs/promises'`.
- `SCOPES` import path is correct for the deeper directory.
- Payload shape, file mode `0o600`, and exit-cleanup are identical to the other two runners.

Worth noting one difference from the other runners: the new block is wrapped in `if (backend === 'fake')`, while the unit and integration runners are unconditional. The conditional is correct here; for `CEP_BACKEND=real` evals, the synthetic cache would be wrong.

Nit, non-blocking: the seeding is now in three runners. Possible follow-up: extract a `test/lib/seed-fake-home.js` helper so the synthetic-cache code is in one place.

TusharKanekiDey:
Thanks Tim,
Will implement the above in a separate follow up PR.

PR #256, opened 2026-05-19
feat(auth): dynamic scope registry with automated integrity enforcement

dylan-klein:
## Summary
This PR establishes a robust **Single Source of Truth (SSOT)** for all OAuth scopes used by the project. It centralizes scope URLs, human-readable names, and functional categories into a single definition block and introduces automated tests to enforce this registry's integrity across all tools.

## Motivation
Previously, OAuth scopes and their metadata were fragmented. Adding a new permission required manual updates to:
1. Hardcoded constants in `lib/constants.js`.
2. Manual documentation strings in `tools/definitions/auth.js`.
3. Informational status reports in `auth_status`.

This fragmentation led to immediate documentation drift and increased the risk of 'shadow scopes' being requested without being properly disclosed to the user.

## Main Changes
- **Unified Registry (`lib/constants.js`):** Introduced `OAUTH_SCOPE_CONFIG` as the authoritative source. The legacy `SCOPES` object and the new `OAUTH_SCOPE_REGISTRY` are now dynamically derived from this single config.
- **Programmatic Documentation (`tools/definitions/auth.js`):** Refactored the `cep_auth` tool to derive its description text by filtering deduplicated categories through a prioritized master list (`SCOPE_CATEGORY_ORDER`).
- **Granular Status Reporting (`cep_auth_status`):** Refactored the status reporter to show individual friendly names (e.g., 'Chrome Reports') rather than raw URLs or broad buckets, providing full transparency into active permissions.
- **Automated Integrity Enforcement (`test/local/tool-registration.test.js`):** Added a new audit suite that programmatically scans every registered tool and asserts that its requested scopes exist in the centralized registry.

### 1. Registry Keyed by URL
**Decision:** The primary lookup registry (`OAUTH_SCOPE_REGISTRY`) is keyed by the raw scope URL string.
**Motivation:** This allows the code to look up metadata for a token's active permissions without needing to know which internal constant ID was used to request them. It handles both constant-based and (accidental) string-based scope requests identically.

### 2. Mandatory Wrapper Exposure
**Decision:** Updated the tool wrapper (`guardedToolCall`) to expose the requested scope array on the returned handler function.
**Motivation:** This was necessary to enable 'Zero-Drift' testing. It allows the test suite to inspect the internal requirements of every tool at registration time without needing to mock the entire Google Auth flow.

### 3. Prioritized Documentation Order
**Decision:** The `cep_auth` tool description follows a canonical order defined in `SCOPE_CATEGORY_ORDER`.
**Motivation:** This restores the high-signal agent-routing order established in #231 (prioritizing Chrome services). A new integrity test ensures every category in the registry is accounted for in this master order, preventing accidental omissions while maintaining the benefits of automation.

### Update (Addressed Reviewer Feedback)
- **Granular Status Reporting:** Fixed a bug in `probe()` and updated the `cep_auth_status` tool to report individual scope names (e.g., "Chrome Reports") instead of broad categories, providing full transparency into active permissions.
- **Prioritized Documentation:** Introduced `SCOPE_CATEGORY_ORDER` to restore the agent-routing signal from #231. The dynamic documentation now strictly follows this prioritized sequence.
- **Strict Integrity Enforcement:** Every registered tool (including the `cep_auth` entry point) is now audited against the registry. Removed all test fallbacks; any tool without explicit scope metadata will now fail the build.


tools/definitions/auth.js
```diff
@@ -36,6 +35,24 @@ const AGENT_HINT =
...
+ * @returns {string} Human-readable summary.
+ */
+function formatAuthStatusSummary(probe) {
+  if (!probe.ok) {
+    return probe.scopesKnown && probe.missingScopes.length > 0
+      ? 'OAuth credentials missing or incomplete permissions.'
+      : 'OAuth credentials not configured.'
+  }
+
+  const granted = probe.grantedScopes || Object.keys(OAUTH_SCOPE_REGISTRY)
```
timfee: I think this falls back to the full registry on every call. `probe()` in `lib/oauth_flow.js` returns an object whose keys are `ok`, `source`, `principal`, `credentialType`, `scopesKnown`, `missingScopes`, `expiry`, and `permissionsWarning`. There's no `grantedScopes` field, so `probe.grantedScopes` is always undefined and the `||` produces `Object.keys(OAUTH_SCOPE_REGISTRY)`. `formatAuthStatusSummary` then returns the full category list for any token, regardless of what was actually granted.

Two ways to fix it, in case you have a preference:
- Add `grantedScopes` to `probe()`'s return value (it's already computed internally as a local `granted`).
- Compute it locally here as `requiredScopes.filter(s => !probe.missingScopes.includes(s))`.

Could you confirm I'm reading `probe()` right?

tools/definitions/auth.js
```diff
@@ -126,9 +145,17 @@ export function registerAuthTools(server, options, sessionState) {
           const requiredScopes = Object.values(SCOPES)
           const cred = oauthFlowCredential({ requiredScopes })
           const probe = await cred.probe()
+
+          // Enhance the probe data with friendly names for the user/agent
+          const statusReport = {
+            ...probe,
+            granted: (probe.grantedScopes || []).map(url => OAUTH_SCOPE_REGISTRY[url]?.name || url),
```
timfee, line 265: Same root cause as line 50. `probe.grantedScopes` is undefined here too, so the `.map()` input is always `[]`, and `statusReport.granted` is empty in every response. Whatever fix you pick above should cover this one as well.

test/local/tool-registration.test.js
```diff
@@ -157,4 +158,44 @@ describe('SEB Tool Registration', () => {
...
+      registerTools(server, {
+        featureFlags: { isEnabled: () => true },
+      })
+
+      const registeredTools = server.registerTool.mock.calls
+
+      for (const call of registeredTools) {
+        const toolName = call.arguments[0]
+        const handler = call.arguments[2]
+        const requestedScopes = handler._scopes || Object.values(SCOPES)
```
timfee: I think this fallback may be making the audit a no-op for `cep_auth`. `cep_auth` is registered without `guardedToolCall` (see the header comment in `auth.js` for why), so `handler._scopes` is undefined on it. The `|| Object.values(SCOPES)` branch then substitutes the full `SCOPES` list, every entry of which is in `OAUTH_SCOPE_REGISTRY` by construction. The "every requested scope exists in the registry" assertion passes for `cep_auth` regardless of what scopes it actually requests.

`cep_auth` is the canonical tool the registry exists to describe, so the test silently passing on it weakens the "Zero-Drift" guarantee from the PR body.

Some options, if useful:
- Drop the `|| Object.values(SCOPES)` fallback and assert `handler._scopes` is defined for every tool. Give `cep_auth` an explicit way to declare its scopes (e.g. a property set during registration).
- Wrap `cep_auth` with `guardedToolCall`, if the pre-flight concern in the header comment can be accommodated some other way.
- Read `cep_auth`'s requested scopes from its definition file directly in the test, separate from the `_scopes` path.

What do you think?

tools/definitions/auth.js
```diff
@@ -53,13 +70,15 @@ export const registerAuthTool = registerAuthTools
...
+  const scopeSummary = getUniqueScopeCategories(Object.values(SCOPES)).join(', ')
+
   server.registerTool(
     TOOL_NAME,
     {
       description:
         'Sign in to Google for the Chrome Enterprise Premium (CEP) MCP server. ' +
         'Use this tool ONLY for the CEP MCP server. The Google Workspace MCP server has its own separate auth tool—do not use this one for that. ' +
-        'Requests the CEP scope set: Chrome browser management, Chrome policy, Cloud Identity (DLP), Admin SDK reports, and Service Usage. ' +
+        `Requests the CEP scope set: ${scopeSummary}. ` +
```
timfee, line 213: The PR body mentions "Greedy Documentation", which I take as the intentional design here, but I wanted to surface a slightly different angle on it. This description string is what agents read to disambiguate the CEP MCP from the Workspace MCP. PR #231 introduced the previous hardcoded string for exactly that purpose, in an importance-ranked order: "Chrome browser management, Chrome policy, Cloud Identity (DLP), Admin SDK reports, and Service Usage".

The new dynamic `scopeSummary` differs in three ways:
1. The order is alphabetical, not importance-ranked.
2. `Identity` and `Licensing` appear in it; they didn't before.
3. `Chrome policy` is grouped under `Chrome browser management` and no longer appears as a separate item.

Structurally, future entries added to `OAUTH_SCOPE_CONFIG` will be reflected here automatically, so the agent-visible description can change without anyone explicitly reviewing the new wording.

Two questions:
- Has anyone re-checked that #231's disambiguation goal still holds with the new wording? (I don't have visibility into how agents have been routing since #231 merged.)
- If preserving the routing signal matters, would you consider keeping the manual string for `cep_auth`'s description and using the registry only for `auth_status` / probe logic? Or adding a `category_order` field to `OAUTH_SCOPE_CONFIG` so the rendered order isn't alphabetical?

timfee (commented):
### Code review

Reviewed at `7745139`. The single-source-of-truth direction looks right to me, and the new integrity test is a good safety net. I have a few questions inline — the first one (around `probe.grantedScopes`) seems like a real bug worth fixing before merge; the others are more "is this what you intended" questions.

One coordination note that isn't tied to a specific line: open PR #232 renames `auth_status` → `cep_auth_status` and `auth_clear` → `cep_auth_clear` in both `tools/definitions/auth.js` and `test/local/tool-registration.test.js`.

tools/definitions/auth.js
dylan-klein: You were right to suspect a fallback—I verified with a reproduction script that `probe()` was indeed missing the `grantedScopes` field. I've added it in `lib/util/credential/oauth_flow.js` and updated the status tool to use it.

I've also switched the status tool summary to report **individual scope names** (e.g., "Admin Audit Reports, Admin Org Units") instead of broad categories. This provides 100% transparency into the current session's permissions and removes any ambiguity regarding "partially granted" categories.

test/local/tool-registration.test.js
dylan-klein: Regarding your point that the `|| Object.values(SCOPES)` fallback makes the audit a no-op for `cep_auth`: you're right that this weakens the "Zero-Drift" guarantee for the tool the registry is most intended to describe.

I've implemented your suggestion to give `cep_auth` an explicit way to declare its scopes. I've manually attached `_scopes` metadata to its handler in `tools/definitions/auth.js`, which allows the integrity test to perform a real audit while keeping the tool itself simple and free of the `guardedToolCall` wrapper (avoiding the pre-flight deadlock).

I've also updated the test in `test/local/tool-registration.test.js` to remove that fallback entirely. Now, every registered tool is required to have explicit `_scopes` metadata; otherwise, the audit will fail.

tools/definitions/auth.js
dylan-klein: To restore the importance-ranked signal established in #231, I've introduced a `SCOPE_CATEGORY_ORDER` constant in `lib/constants.js`. The dynamic tool description now filters the registry categories through this list to ensure "Chrome policy" and "Chrome browser management" always appear first.

I've added an integrity test that asserts `SCOPE_CATEGORY_ORDER` contains **every** category defined in the registry. This preserves the SSOT merits: it's no longer possible to add a new scope category without also explicitly defining its presentation priority, ensuring the documentation remains both automated and intentional.

tools/definitions/auth.js
dylan-klein: Acknowledged. I've added the missing `grantedScopes` field to `probe()`. This ensures both the status summary and this `statusReport.granted` list are now correctly populated with actual permissions.

PR #257, opened 2026-05-19
feat(auth): dynamic environment-aware agent hints and user instructions

dylan-klein:
## Summary
This PR implements dynamic, context-sensitive instructions for both users and AI agents during the OAuth sign-in flow. It distinguishes between workstation (local) and headless (SSH/remote) environments to reduce user friction.

## Motivation
Previously, the server provided a single set of 'paste-back' instructions regardless of the environment. This confused local workstation users (where the browser opened automatically) and failed to provide clear guidance for remote users on how to handle the '127.0.0.1 Connection Refused' redirect.

## Main Changes
- **Dynamic Hints:** Refactored `cep_auth` to select between `AGENT_HINT_WORKSTATION` and `AGENT_HINT_HEADLESS` based on the `browserOpened` flag.
- **Improved User Guidance:** Updated the `awaitingResponse` text to provide a technically accurate, step-by-step guide for manual sign-in in headless environments.
- **Agent Knowledge:** Added explicit instructions for the agent on how to guide the user in the workstation case (sign in and close tab).

### 1. Tool-Driven Branching
**Decision:** The branching logic lives in the tool definition (`auth.js`) rather than the core auth logic.
**Motivation:** This keeps the instructions close to the presentation layer (MCP tool response) while allowing the core logic to remain purely functional.

### 2. Matching Agent and User Text
**Decision:** The instructions provided to the agent in the `agentHint` field mirror the steps shown to the user in the `content` field.
**Motivation:** This ensures the agent and user are always 'on the same page' when troubleshooting or following the sign-in flow.

## Verification
- **Unit Tests:** `test/local/auth-tool.test.js` passes 100% with updated assertions.
- **Manual Verification:** Confirmed that the output text correctly reflects the intended scenarios.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -7,3 +7,5 @@ results_final.json
 .gemini/
 .opencode/
 *.patch
+.venv/
+.stryker-tmp/
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -41,7 +41,17 @@ const copyrightHeader = [
 
 export default [
   {
-    ignores: ['**/dist', '**/node_modules', 'results/**', '.worktrees/**', '.claude/**', '.gemini/**', '.opencode/**'],
+    ignores: [
+      '**/dist',
+      '**/node_modules',
+      'results/**',
+      '.worktrees/**',
+      '.claude/**',
+      '.gemini/**',
+      '.opencode/**',
+      '**/.venv/**',
+      '**/.stryker-tmp/**',
+    ],
   },
   js.configs.recommended,
   nodePlugin.configs['flat/recommended'],
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -28,37 +28,84 @@ export const SERVICE_NAMES = {
   SERVICE_MANAGEMENT: 'servicemanagement.googleapis.com',
 }
 
-export const SCOPES = {
-  OPENID: 'openid',
-  EMAIL: 'https://www.googleapis.com/auth/userinfo.email',
-  CHROME_MANAGEMENT_POLICY: 'https://www.googleapis.com/auth/chrome.management.policy',
-  CHROME_MANAGEMENT_REPORTS_READONLY: 'https://www.googleapis.com/auth/chrome.management.reports.readonly',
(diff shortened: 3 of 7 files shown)


timfee (approved):
Two things to flag:

**1. `probe.grantedScopes` is never set, so the auth-status summary is wrong on every successful probe.** `CredentialProbe` only carries `missingScopes` (see the typedef and the only implementation in `lib/util/credential/oauth_flow.js`), so the fallback `probe.grantedScopes || Object.keys(OAUTH_SCOPE_REGISTRY)` always resolves to *every* registered scope. Any signed-in user gets "Authorized for: Admin SDK reports, Chrome browser management, …" regardless of what was actually consented, and `statusReport.granted` is always `[]`. Either derive `granted = requiredScopes - missingScopes` when `scopesKnown`, or extend `oauthFlowCredential.probe()` to emit `grantedScopes` from `tokens.scope`.

https://github.com/google/chrome-enterprise-premium-mcp/blob/6c41a37b11498149b9083f396315cde9a54a4565/tools/definitions/auth.js#L50-L65
https://github.com/google/chrome-enterprise-premium-mcp/blob/6c41a37b11498149b9083f396315cde9a54a4565/tools/definitions/auth.js#L155-L165

**2. Branch is behind main — merging as-is silently reverts several recently-landed auth features.** This PR branched before #232/#244/#258/#259/#260. Merging as-is would undo:

- tool names `cep_auth_status` / `cep_auth_clear` revert to `auth_status` / `auth_clear` (the `cep_` prefix was added in #232 to disambiguate from Workspace MCP)
- the OSC 8 hyperlink block from #244 is deleted from `awaitingResponse`
- the `cliFallbackLine(...)` (promoting `npx … auth login` as the primary sign-in path, added in #259) is deleted
- the longer `cep_auth_status` / `cep_auth_clear` descriptions from #232 revert to their pre-#232 one-liners

https://github.com/google/chrome-enterprise-premium-mcp/blob/6c41a37b11498149b9083f396315cde9a54a4565/tools/definitions/auth.js#L137-L196
https://github.com/google/chrome-enterprise-premium-mcp/blob/6c41a37b11498149b9083f396315cde9a54a4565/tools/definitions/auth.js#L222-L262

A rebase onto current main will surface these as conflicts so they're resolved deliberately.

PR #258, opened 2026-05-19
fix(auth): prevent mangled consent URLs by conditionally wrapping in OSC 8

TusharKanekiDey:
Prevent mangled consent URLs and duplicate escape characters in terminals/panels that do not support OSC 8 hyperlinks.

### Description
This PR conditionally formats the Google OAuth consent URL printed by the `cep_auth` tool. It resolves issue b/514456098.

When the terminal environment supports OSC 8 hyperlinks:
- Displays a short, wrap-resilient hyperlink (`🔗 Click here to open the Google Sign-in page in your browser`) that avoids layout wrapping and vertical borders (`│`) mangling the URL.
- Also prints the clean plain URL as a backup.

When the terminal/client does not support OSC 8 hyperlinks:
- Prints **only** the plain, clean URL.
- Completely removes all raw escape characters and duplicate URLs, significantly simplifying manual copy-pasting and cleanup.

### Changes
- Added the `supportsHyperlinks()` helper in `tools/definitions/auth.js` to check standard environment signals.
- Updated `osc8()` helper to support custom labels.
- Refactored `awaitingResponse()` to conditionally output short hyperlinks or plain fallbacks.
- Added comprehensive unit tests in `test/local/auth-tool.test.js` covering both supported and unsupported hyperlink environments.

### DIFF
--- a/test/local/auth-tool.test.js
+++ b/test/local/auth-tool.test.js
@@ -64,37 +64,79 @@ describe('cep_auth Tool', () => {
     assert.strictEqual(completeToolAuth.mock.callCount(), 0)
   })
 
-  test('When cep_auth is invoked and the wait window expires, then it returns status=awaiting with the consent URL and paste-redirect-url next action', async () => {
-    const startToolAuth = mock.fn(async () => ({
-      status: 'awaiting',
-      authUrl: 'https://accounts.google.com/o/oauth2/v2/auth?state=ABC',
-      browserAttempted: false,
-      browserOpened: false,
-      expiresAt: new Date(Date.now() + 300_000),
-      source: 'managed',
-    }))
-    const completeToolAuth = mock.fn()
-    await register({ startToolAuth, completeToolAuth })
+  test('When cep_auth is invoked and the wait window expires with terminal hyperlink support, then it returns status=awaiting with OSC 8 hyperlink and plain URL', async () => {
+    process.env.FORCE_HYPERLINK = '1'
+    try {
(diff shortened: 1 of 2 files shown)


timfee:
#244 assumed that what reaches the terminal still contains OSC 8 escapes that get rendered there. That assumption holds for CLIs attached directly to a TTY (gcloud, firebase, npm login). It does not hold for an MCP server: server output is rendered first by the client UI, and only what survives that pass reaches the terminal. In the reported case the client is Gemini CLI, where the text block is displayed in a bordered panel and OSC 8 is not processed. So the raw `]8;;...\\` bytes were displayed, and the URL was shown twice (once as link target, once as visible label).

This PR's `supportsHyperlinks()` check is almost certainly enough for the reported case. `TERM_PROGRAM` / `VTE_VERSION` are not usually inherited by MCP child processes from the launching client, so the plain-URL fallback path gets taken—which is the right outcome.

The detection is a heuristic, though. The question being answered is "is the host terminal modern?" when the question that actually matters is "does the MCP client's text renderer pass OSC 8 through?" The two questions can have different answers. For the reported case those answers happen to agree.

Worth landing as-is. Filed #259 as a docs-only follow-up promoting the existing `npx ... auth login` CLI as the primary sign-in path, so first-time sign-in does not depend on how a client renders the URL.

timfee (approved):

TusharKanekiDey:
Sounds good.
Merging this PR.
But we can go with #259 as long run solution.
Thanks for the PR

PR #259, opened 2026-05-19
promote npx auth CLI as the primary sign-in path; surface it from cep_auth

timfee:
## Summary

Two related changes that route users to the CLI sign-in when the in-agent consent URL is unreliable.

- README sign-in section: show the explicit `npx @google/chrome-enterprise-premium-mcp auth login` command, frame it as the recommended first-time sign-in path, note the `cep_auth` rendering caveat.
- `cep_auth` awaiting response: append a line pointing at the same CLI command, so users who hit the broken render in their MCP client see the escape hatch in the very response that is broken.

The CLI line is conditional on the resolved OAuth client source. For the bundled managed client, the bare `npx ... auth login` works. For a custom client (`CEP_OAUTH_CLIENT_ID` / `CEP_OAUTH_CLIENT_SECRET` set on the server), the user must also export those env vars in their shell, otherwise the CLI would silently sign in under the bundled managed client and cache a token under the wrong identity.

## Context

Follow-up from #244 / #258. The `cep_auth` tool returns the consent URL inside an MCP text block. Some clients (Gemini CLI's panel renderer in the reported case) display the text block inside a Unicode-bordered panel where OSC 8 escape sequences are not processed, so the URL renders as raw escape bytes or gets clipped at the border.

#244 wrapped the URL in OSC 8 escapes on the assumption that the terminal would render them. The client-side panel sits in front of the terminal and strips the escapes, so the assumption did not hold.

#258 conditionally emits OSC 8 only when `process.env` signals a hyperlink-capable terminal. That fixes the reported case in practice because MCP child processes typically do not inherit `TERM_PROGRAM` etc., so the heuristic falls back to plain URL. The detection is a heuristic, though, and can misfire when env vars do propagate to the child but the client panel still strips escapes.

The CLI sidesteps the question. The user runs it in their shell, the URL is printed straight to the terminal, and a token is cached at `~/.config/cep-mcp/tokens.json` for the server to read on every call. No client-side rendering is in the loop.

#258 is still worth landing as the polite fallback when `cep_auth` does get invoked. This PR makes the CLI the recommended path, and makes sure users who do hit `cep_auth` see how to recover from a bad render.

## Files

- `README.md` — sign-in section rewritten
- `tools/definitions/auth.js` — adds `cliFallbackLine(source)` and appends it to the awaiting response; uses `resolveOAuthClientConfig().source` to choose between managed and custom variants
- `test/local/auth-tool.test.js` — adds `withClientEnv` helper plus two BDD tests covering the managed and custom variants

## Test plan

- [x] Unit suite (`npm run test:unit`) — 454 passing
- [x] ESLint clean on the touched files
- [x] Prettier clean on the touched files
- [ ] After #258 merges, rebase to confirm no functional conflict (the diffs touch different regions of `awaitingResponse`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,17 +21,26 @@ npm install
 
 ### 1. Sign in
 
-Run `mcp auth login` and approve consent in the browser. The CLI caches
-the access token at `~/.config/cep-mcp/tokens.json`.
+Run the auth CLI and approve consent in your browser:
 
-If a tool call runs and the cached token has expired, the agent will
-offer to sign you in. It does this by calling the `cep_auth` tool, which
-opens a browser tab for consent and stores the new token in the same
-place.
+```bash
+npx @google/chrome-enterprise-premium-mcp auth login
+```
+
+The access token is cached at `~/.config/cep-mcp/tokens.json`. From a
+local checkout you can also run `npm run auth:login`.
+
+Prefer this path for the initial sign-in. The consent URL is printed by
+your shell, so OSC 8 hyperlinks and ordinary URL selection both work.
+
+You can also start sign-in from inside the agent using the `cep_auth`
+MCP tool. Some MCP clients render text inside bordered panels that strip
+terminal escape sequences, which can mangle the consent URL. If that
+happens, fall back to the CLI command above.
 
 To use a custom OAuth client in a Cloud project of your own, set
-`CEP_OAUTH_CLIENT_ID` and `CEP_OAUTH_CLIENT_SECRET` and re-run
-`mcp auth login`. See
+`CEP_OAUTH_CLIENT_ID` and `CEP_OAUTH_CLIENT_SECRET` and re-run the auth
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -21,17 +21,26 @@ npm install
 
 ### 1. Sign in
 
-Run `mcp auth login` and approve consent in the browser. The CLI caches
-the access token at `~/.config/cep-mcp/tokens.json`.
+Run the auth CLI and approve consent in your browser:
```
TusharKanekiDey, line 24: Thanks for updating the readme.

TusharKanekiDey (approved):

dylan-klein (approved):
Approved. Note: While the PR description suggests promoting the CLI as the primary path, the implementation in tools/definitions/auth.js correctly treats it as a fallback ('If the URL above is hard to copy...').

PR #265, opened 2026-05-20
fix(auth): dim plain URL fallback inside TUI container

TusharKanekiDey:
Strip protocol prefix from plain fallback consent URL to deactivate terminal native hyperlink parsing.

### Description
This PR strips the `https://` or `http://` protocol prefix from the plain fallback consent URL printed inside the boxed TUI panel.

This deactivates native terminal/CLI link parsing for this line (meaning it will not be clickable or highlighted), preventing accidental clicks on wrapped or mangled URL fragments inside boxed TUI containers. The fallback URL remains fully copy-pasteable (browsers automatically resolve protocol-less URLs), and the primary "Sign in to Chrome Enterprise Premium" markdown link printed below remains standard and fully highlighted.

Instruction texts have also been updated to explicitly mention "fallback URL" and "copy-paste only".

### Changes
- Stripped the `https?://` prefix from `result.authUrl` when printing the fallback URL in `awaitingResponse()`.
- Updated surrounding instruction texts to guide users toward copy-pasting.
- Updated assertions in `test/local/auth-tool.test.js` to resiliently search for the protocol-less URL using `.includes()`.

### DIFF
--- a/test/local/auth-tool.test.js
+++ b/test/local/auth-tool.test.js
@@ -88,7 +88,7 @@ describe('cep_auth Tool', () => {
       assert.match(result.content[0].text, /accounts\.google\.com/)
 
       const lines = result.content[0].text.split('\n')
-      const plainUrlIndex = lines.findIndex(l => l.includes('https://accounts.google.com/o/oauth2/v2/auth?state=ABC'))
+      const plainUrlIndex = lines.findIndex(l => l.includes('accounts.google.com/o/oauth2/v2/auth?state=ABC'))
       assert.ok(plainUrlIndex > 0, 'plainUrl should appear in the text block')
       assert.strictEqual(lines[plainUrlIndex - 1], '', 'plainUrl should have a blank line above it')
       assert.strictEqual(lines[plainUrlIndex + 1], '', 'plainUrl should have a blank line below it')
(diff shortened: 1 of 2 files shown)


TusharKanekiDey:
Sample screenshot - https://screenshot.googleplex.com/UpE87UeRCL46UDs

timfee (approved):
Diff doesn't match the PR description — the description says this wraps the URL in ANSI dim via a new `dim()` helper, but the actual change strips `https://` from the printed string:

https://github.com/google/chrome-enterprise-premium-mcp/blob/8e6ae9f28d4fc760d4b75ebfb941695bd5ebeaee/tools/definitions/auth.js#L267-L271

The surrounding text still says "Copy and paste this fallback URL into a browser" but the printed value is no longer a URL — many browsers/launchers treat scheme-less strings as search queries, regressing #258's clean copy-paste fix.

https://github.com/google/chrome-enterprise-premium-mcp/blob/8e6ae9f28d4fc760d4b75ebfb941695bd5ebeaee/tools/definitions/auth.js#L259-L271

The test assertions were also relaxed from `=== 'https://…'` to `.includes('accounts.google.com/…')`, which passes against the scheme-stripped output and hides the regression:

https://github.com/google/chrome-enterprise-premium-mcp/blob/8e6ae9f28d4fc760d4b75ebfb941695bd5ebeaee/test/local/auth-tool.test.js#L86-L92

Could you confirm intent — implement the `dim()` helper as the description says, or revisit the scheme-stripping approach?

dylan-klein (approved):

PR #270, opened 2026-05-23
fix: secure default binding for HTTP server (CWE-1188)

islamdatabridge-cpu:
This PR addresses the insecure default binding in HTTP mode as reported in IssueTracker 516028420.


islamdatabridge-cpu:
@googlebot I signed it!

islamdatabridge-cpu:
@googlebot check

PR #274, opened 2026-05-26
fix: support /apps prefix for fake licensing user endpoint

TusharKanekiDey:
Resolves the licensing evaluation failures (l01, i05) caused by 404 path mismatches on the fake API server.

### DIFF
--- a/test/helpers/fake-api-server.js
+++ b/test/helpers/fake-api-server.js
@@ -364,26 +364,32 @@ export function createFakeApp() {
   )
 
   // Licensing: Get User License
-  app.get('/licensing/v1/product/:productId/sku/:skuId/user/:userId', (req, res) => {
-    for (const customerLicenses of Object.values(state.licenses)) {
-      const skuLicenses = customerLicenses[req.params.productId]?.[req.params.skuId] || []
-      const license = skuLicenses.find(l => l.userId === req.params.userId)
-      if (license) {
-        // Return a structure matching the real Google Licensing API single response
-        return res.json({
-          kind: 'licensing#licenseAssignment',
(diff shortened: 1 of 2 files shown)


dylan-klein (approved):
### 1. SUMMARY
Added support for the `/apps` path prefix in the fake API server's licensing endpoint (mapping both `/licensing/v1/product/...` and `/apps/licensing/v1/product/...` to the same handler). Added a unit test in `test/local/fake-api-server.test.js` to verify that requests using the `/apps` prefix are handled correctly.

This aligns the fake API server with the real Google Licensing API behavior (which uses the `/apps` prefix, as already reflected in the mock `selfLink` response).

### 2. RISK ASSESSMENT
*   **Systemic Risk**: Low. Changes are strictly inside test helpers (`fake-api-server.js`) and unit tests. No production code is affected.
*   **Operational Risk**: Low. Fully backward-compatible change. Using an array of paths in Express router is standard and safe.
*   **Security Risk**: Low. No security-sensitive code or data touched.

### 3. DESIGN EVALUATION
The design is clean and highly maintainable. Using Express path arrays `app.get([path1, path2], handler)` is the standard way to support multiple routes under the same logic without duplicating the handler code. This ensures the fake API server remains realistic and consistent with the `selfLink` it returns.

### 4. ACTIONABLE SUGGESTIONS
| Priority | Suggestion | Rationale |
|---|---|---|
| **Needed (Blocking)** | None | |
| **Possible (Nit)** | Add a 404 test case for the `/apps` prefix path in `test/local/fake-api-server.test.js` | While it uses the same handler as the non-prefixed path, explicitly verifying that unknown users return a 404 on the new path would increase test coverage and robustness. |

PR #299, opened 2026-06-01
refactor(auth): make non-headless authentication non-blocking

dylan-klein:
### Motivation
In the non-headless authentication path, calling the `cep_auth` tool opened the user's browser but synchronously blocked (hung) the terminal for up to 90 seconds waiting for the loopback callback. Because the tool call was blocked, the user received no in-terminal feedback, often missing the browser tab and wondering why the server was frozen. Furthermore, standard tool pre-flights tried to trigger this auto-login inline, causing standard operations to hang silently on missing credentials.

### Main Changes
1. **Non-blocking `cep_auth` (`lib/util/credential/auth_login.js` & `tools/definitions/auth.js`)**:
   - Removed the synchronous blocking wait (`Promise.race`) from `startToolAuth`. It now returns `status: awaiting` **instantly** with the backup URL and instructions.
   - Implemented a background promise on the loopback server that listens for the OAuth redirect asynchronously. When the user signs in, the background handler automatically exchanges the authorization code, caches the token, and shuts down the loopback server without requiring manual action.
   - Removed the over-engineered OSC 8 terminal hyperlinking that caused ugly escaped characters (`\u001b]8;;...`) in several terminal clients. It now outputs a single plain, highly-compatible URL which modern terminals naturally make clickable.
2. **Fail-Fast Pre-flights (`tools/utils/wrapper.js`)**:
   - Updated `guardedToolCall` to fail-fast if the OAuth token is invalid or missing, returning a clean remediation error immediately instead of proceeding.
3. **Removed Inline Auto-Login (`lib/util/auth.js`)**:
   - Removed the inline `startToolAuth` auto-login attempt inside `getAuthClient` for stdio mode to prevent standard tools from ever hanging the transport.

### Non-Trivial Design Decisions
We evaluated three viable paths and chose **Option 1 (Non-blocking Browser Flow with Background Loopback & Paste-back Fallback)** over:
*   **Option 2 (Pure Headless-like Flow)**: Disabling the loopback server and forcing manual URL copy-pasting even in desktop environments. We rejected this as it degrades the user experience.
*   **Option 3 (Short Wait)**: Simply reducing the timeout to e.g., 2 seconds. We rejected this as it still causes a brief freeze and feels hacky.

Option 1 preserves the seamless "magic" of automatic loopback token caching while completely solving the terminal freeze by returning immediately. We also chose to completely remove OSC 8 ANSI hyperlinking in favor of plain URLs, since JSON-serialization inside the MCP transport often escapes ANSI sequences (rendering raw `\u001b` sequences) and modern terminals natively handle plain URL linkification anyway.

### Testing Verified
* All unit tests passed (**476/476**).
* Modified local auth tests passed (**30/30**), covering the new background callback, mismatched state, and `access_denied` handling.
* All fake integration tests (**28/28**) and smoke tests passed successfully.

TAG=agy
CONV=c4bb9311-f91c-4724-a989-bbda87308026


TusharKanekiDey:
Is it possible to add test for this change ?

dylan-klein:
@TusharKanekiDey Done. Added a dedicated unit test to `test/local/auth_login.test.js` (`When a loopback callback fires after the pending session has been cleared or reset, it does not write the cache or proceed`) to assert that old/stale loopback callbacks exit early and do not write any tokens if the active pending auth session has been cleared or restarted in the meantime.

TusharKanekiDey (approved):

PR #301, opened 2026-06-02
feat(auth): fail early on missing credentials instead of opening browser

dylan-klein:
feat(auth): fail early on missing credentials instead of opening browser

## Motivation

Currently, when the MCP server is in stdio mode and has no valid cached credentials, it attempts to open a browser inline during a tool call. If the user misses this (e.g., it opens in the background or on another workspace), the tool call blocks indefinitely waiting for consent. This leaves the user with an apparently frozen interface and no feedback that they need to take action in a browser.

This PR refactors the flow to fail early, directing the user to run the explicit `cep_auth` tool (which will open the browser if possible) or use the CLI to authenticate.

### Core
1. **lib/util/auth.js** (`getAuthClient`):
   - Removed the inline browser authentication flow (`startToolAuth`) when in stdio mode.
   - Now throws an `Authentication required` error immediately, directing the user to run `cep_auth` or `auth login`.
2. **tools/utils/wrapper.js** (`guardedToolCall`):
   - Updated the pre-flight check to fail early if the token is invalid, regardless of browser availability (previously it only failed early in headless mode).
   - Added a `skipAuthCheck` option to allow specific tools to bypass this check.

### Tools
1. **tools/definitions/auth.js**:
   - Applied `skipAuthCheck: true` to `cep_auth_status` and `cep_auth_clear` so they can still be used to check status or clear cache even when unauthenticated.

### Tests
1. **test/local/auth.test.js**:
   - Added a test to verify `getAuthClient` throws immediately in stdio mode when tokens are missing.
2. **test/local/utils.test.js**:
   - Added a test to verify `skipAuthCheck: true` successfully bypasses the pre-flight check.
   - Updated existing early-failure tests to remove `SSH_CONNECTION` forcing, verifying they fail early in all environments.

### DIFF
--- a/lib/knowledge/11-dlp-rule-reference.doc.js
+++ b/lib/knowledge/11-dlp-rule-reference.doc.js
@@ -20,6 +20,6 @@ export const doc = {
   title: 'Chrome DLP Rule Configuration Reference',
   articleId: '11',
   summary:
-    'Comprehensive technical reference for authoring Chrome DLP rules. Helps with CEL condition logic and action configuration. Covers 100+ Data Type Detectors, 286+ Web Categories, and trigger compatibility constraints. Keywords: matches_dlp_detector, url_category, CEL syntax, AUDIT/WARN/BLOCK behavior, MIME types.',
+    'Comprehensive technical reference for authoring Chrome DLP rules. Helps with CEL condition logic and action configuration. Covers 100+ Predefined Detectors, 286+ Web Categories, and trigger compatibility constraints. Keywords: matches_dlp_detector, url_category, CEL syntax, AUDIT/WARN/BLOCK behavior, MIME types.',
   content: generateDlpCelReference(),
 }
--- a/lib/knowledge/11-dlp-rule-reference.doc.js
+++ b/lib/knowledge/11-dlp-rule-reference.doc.js
@@ -20,6 +20,6 @@ export const doc = {
   title: 'Chrome DLP Rule Configuration Reference',
   articleId: '11',
   summary:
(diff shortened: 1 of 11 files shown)


TusharKanekiDey:
Please fix the failing unit test

lib/knowledge/11-dlp-rule-reference.doc.js
```diff
@@ -20,6 +20,6 @@ export const doc = {
   title: 'Chrome DLP Rule Configuration Reference',
   articleId: '11',
   summary:
-    'Comprehensive technical reference for authoring Chrome DLP rules. Helps with CEL condition logic and action configuration. Covers 100+ Predefined Detectors, 286+ Web Categories, and trigger compatibility constraints. Keywords: matches_dlp_detector, url_category, CEL syntax, AUDIT/WARN/BLOCK behavior, MIME types.',
+    'Comprehensive technical reference for authoring Chrome DLP rules. Helps with CEL condition logic and action configuration. Covers 100+ Data Type Detectors, 286+ Web Categories, and trigger compatibility constraints. Keywords: matches_dlp_detector, url_category, CEL syntax, AUDIT/WARN/BLOCK behavior, MIME types.',
```
TusharKanekiDey: Did not understand why changed the name.
Although I am cool with either of them

dylan-klein: I've reverted this change back to 'Predefined Detectors' in the latest push. Reverting to the generic term is better since it covers both data type detectors and masking detectors.

dylan-klein:
I've fixed the unit test failure in the latest push. The issue was that the test was sensitive to GOOGLE_APPLICATION_CREDENTIALS being set in the environment (which happens in CI but not always locally). I've updated the test to explicitly unset it during execution so it runs consistently.

TusharKanekiDey (approved):

PR #303, opened 2026-06-02
feat(auth): implement robust instant asynchronous auth flow for cep_auth

dylan-klein:
feat(auth): implement robust instant asynchronous auth flow for cep_auth

**Depends on #301**

## Motivation

During testing of the "fail early" auth refactor (#301), we identified that the explicit `cep_auth` tool itself could block silently for up to 90 seconds if it attempted to open a browser. If the user misses the browser tab (e.g., it opens in the background or on a remote workstation screen they cannot see), the tool call blocks indefinitely. This causes client-side timeouts in LLM frameworks/clients (which often have 30s timeouts) and leaves the user with a "hung" terminal experience.

This PR refactors the `cep_auth` tool flow to be **fully instant** (0ms wait time) and asynchronous, completely eliminating silent hangs while maintaining robust auto-login capabilities.

### Core
1. **lib/util/credential/auth_login.js** (`startToolAuth` & `completeToolAuth`):
   - **Instant Return**: Removed all synchronous waiting from `startToolAuth` (0ms wait). The tool call now launches the browser and immediately returns `status: 'awaiting'` with instructions and the URL, providing instant feedback.
   - **Background Token Exchange**: Refactored the loopback callback to run in the background. Even though `startToolAuth` returns immediately, the background loopback server continues running for up to 5 minutes. If the user signs in, the server automatically exchanges the code and caches the token in the background.
   - **Robust Completion**: Updated `completeToolAuth` to check local token validity first. If the background flow already cached the token (same-machine), pasting the redirect URL will now immediately return success instead of throwing a `NO_PENDING_AUTH` error. For remote machines (where loopback fails), the user pastes the redirect URL, which is successfully exchanged via the active pending auth.

### Tools
1. **tools/definitions/auth.js**:
   - **Agent Warnings**: Updated the `cep_auth` tool description to explicitly instruct the AI agent to **warn the user and ask for confirmation** before calling the tool. This ensures the user expects the browser launch and is ready to sign in.

### Tests
1. **test/local/auth_login.test.js**:
   - **Test Isolation**: Fixed test isolation leaks by ensuring `completeToolAuth` tests use an isolated temporary `cachePath` instead of reading/writing to your real local token cache at `~/.config/cep-mcp/tokens.json`.
   - **Cleaned Up Legacy Tests**: Removed legacy synchronous wait/timeout tests that are no longer valid under the instant-return design.
   - **Background Exchange Test**: Added a unit test to verify that `startToolAuth` returns `awaiting` immediately and the background flow successfully exchanges and caches the token when the callback fires.
   - **Background Mismatch & Error Tests**: Added unit tests to verify that background state mismatches or `access_denied` callbacks correctly clean up resources (clear `pendingAuth` and stop the server).
   - **Robust Completion Test**: Added a unit test to verify `completeToolAuth` returns success immediately if a valid token is already cached.


TusharKanekiDey (commented):
I have reviewed this PR and it looks great! All 494 tests are passing, and evals validated successfully in dry-run.

### Review Summary
- **UX Improvements**: The OSC 8 hyperlink support and cleaner instruction logic (concise when browser opens, detailed manual steps when it doesn't) are excellent UX additions.
- **Safety**: Adding the warning in `cep_auth` description is a crucial safety guardrail for the agent.
- **Robustness**: The early return in `completeToolAuth` when a cached token is already valid is a good robustness improvement.
- **Flexibility**: The `authMethod` arg provides good control for headless or automation scenarios.

TusharKanekiDey (approved):

PR #314, opened 2026-06-04
feat(chrome_management): add security insights query methods to ChromeManagementClient

sahil-madeka:
# Implementation Summary: Security Insights Query Methods

This change adds wrapper methods to the `ChromeManagementClient` for querying Chrome Enterprise Security Insights data, specifically content transfers and URL visits, including their breakdowns. These methods use the `v1alpha1` version of the API as documented. These methods will be used by a new tool call security_insights_data_tool whose registration will be guarded by an experiment flag

#### [chrome_management_client.js](file:///usr/local/google/home/sahilmadeka/.gemini/jetski/worktrees/chrome-enterprise-premium-mcp/run-latest-commit-tests/lib/api/chrome_management_client.js)

*   **`_getSecurityInsightsUrl(resolvedId, action, apiVersion = 'v1')`**: Updated to support a dynamic `apiVersion` parameter to allow calling `v1alpha1` endpoints.
*   **`queryContentTransfers(customerId, options, authToken)`**: Calls `queryContentTransfers` (v1alpha1) to get high-level content transfer summaries.
*   **`queryContentTransfersBreakdowns(customerId, options, authToken)`**: Calls `queryContentTransfersBreakdowns` (v1alpha1) to get detailed breakdowns by user, domain, or category.
*   **`queryUrlVisits(customerId, options, authToken)`**: Calls `queryUrlVisits` (v1alpha1) to get high-level URL visit summaries.
*   **`queryUrlVisitsBreakdowns(customerId, options, authToken)`**: Calls `queryUrlVisitsBreakdowns` (v1alpha1) to get detailed breakdowns by user or domain.

#### [chromemanagement.test.js](file:///usr/local/google/home/sahilmadeka/.gemini/jetski/worktrees/chrome-enterprise-premium-mcp/run-latest-commit-tests/test/local/chromemanagement.test.js)

*   Added unit tests for each of the new methods to ensure:
    *   The correct `v1alpha1` URL is constructed.
    *   Query parameters (options) are correctly passed.
    *   The authorization token is correctly threaded.

## Verification Results

All unit tests and integration tests passed successfully during presubmit.

PR #317, opened 2026-06-09
docs: add Querying Security Insights Data knowledge article

sahil-madeka:
This PR adds a knowledge document for the Query insights data API. The public doc is currently ACLd so we need to provide a body describing the APIs

### DIFF
--- a/lib/knowledge/31-security-insights-data.md
+++ b/lib/knowledge/31-security-insights-data.md
@@ -1,0 +1,41 @@
+---
+title: Querying Security Insights Data
+articleId: '31'
+summary: 'Guide to querying Chrome Enterprise Security Insights data. Covers methods for retrieving summaries and breakdowns of content transfers and URL visits. Helps with understanding security posture and data movement.'
+---
+
+# Querying Security Insights Data
+
+The `SecurityInsightsService` `Query*` methods provide aggregated data about user activities related to content movement and website visits. The agent can analyze this data to identify potential security risks and areas where Data Loss Prevention (DLP) rules could be effectively applied to enhance data protection.
+
+## 1. QueryContentTransfers
+
+- **Insight for DLP**: Provides the agent with a high-level, organization-wide summary of content transfer volumes (uploads, downloads, prints), distinguishing between total and sensitive data.
+- **Use Case**: The agent can use this data to baseline data movement and flag significant increases in sensitive data transfers, indicating a potential need for broader DLP controls.
+- **DLP Suggestion**: If the agent observes high sensitive data transfer volumes, it can suggest implementing foundational DLP rules, such as auditing or warning on the transfer of data matching common sensitive information types.
+
+## 2. QueryContentTransfersBreakdowns
+
+- **Insight for DLP**: Allows the agent to perform a granular analysis of content transfers, broken down by User, Event Domain, or Content Category for a specific metric.
+- **Use Case**: The agent can pinpoint specific sources of risk:
+  - **USER**: The agent can identify individuals or roles with high-volume sensitive data transfers.
+  - **EVENT_DOMAIN**: The agent can highlight external domains frequently receiving sensitive company data.
+  - **CONTENT_CATEGORY**: The agent can determine which types of sensitive information are most often transferred.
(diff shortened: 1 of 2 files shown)


dylan-klein (commented):
I noticed the documentation uses phrasing like "Provides the agent with..." and "The agent can use this data...". Since these tools aren't currently registered in the MCP server (PR #314 adds them to the API client, but doesn't seem to expose them as tools), the agent won't have direct access to call them. 

Are these intended to be exposed as MCP tools in this PR (or a companion PR)?  If so, we could stack the PRs to manage this dependency cleanly and prevent the agent from trying to call non-existent tools:

1. **PR #314 (API Client)** first.
2. **A new PR (Tool Registration)** stacked on top of PR #314.
3. **This PR #317 (Docs)** rebased to stack on top of the Tool Registration PR.

Additionally, it will be easier to evaluate the info this PR adds when we can see what info is provided in tool descriptions themselves. Presumably, tool descriptions will contain some info on how to call the tools.

lib/knowledge/31-security-insights-data.md
```diff
@@ -0,0 +1,41 @@
+---
+title: Querying Security Insights Data
+articleId: '31'
+summary: 'Guide to querying Chrome Enterprise Security Insights data. Covers methods for retrieving summaries and breakdowns of content transfers and URL visits. Helps with understanding security posture and data movement.'
+---
+
+# Querying Security Insights Data
+
+The `SecurityInsightsService` `Query*` methods provide aggregated data about user activities related to content movement and website visits. The agent can analyze this data to identify potential security risks and areas where Data Loss Prevention (DLP) rules could be effectively applied to enhance data protection.
```
dylan-klein, line 9: Now that PR #331 (`feat: implement security_insights_data tool and tests`) has merged into `main`, the prerequisite tools are live in the MCP server under the tool name **`security_insights_data`**.

However, because this article references API-level method names (`QueryContentTransfers`, `QueryUrlVisits`) without mentioning the actual MCP tool name or parameter schema, an agent reading this document will search its available tool list for a tool named `QueryContentTransfers`. When it doesn't find it, it assumes it lacks the capability.

To align with how other knowledge docs guide the agent (e.g., `12-security-posture-guide.md` explicitly citing `get_chrome_activity_log`), please update the introduction and headings to explicitly cite the tool name and action parameters.

**Suggested wording for the intro:**
```markdown
To query this data, call the `security_insights_data` tool with the corresponding query name as the `action` parameter (e.g., `action: 'queryContentTransfers'`).
```

dylan-klein:
FYI, i15 and i16 evals do not pass. We currently have periodic checks for evals, but no presubmit checks. I do think we need presubmit checks.

sahil-madeka:
> FYI, i15 and i16 evals do not pass. We currently have periodic checks for evals, but no presubmit checks. I do think we need presubmit checks.

https://github.com/google/chrome-enterprise-premium-mcp/pull/359 are we still having issues after merging that PR @dylan-klein ?

dylan-klein:
PR #359 enabled the tool flag and fixed the prompt structure so the agent now successfully calls `security_insights_data`.

However, multi-run testing shows `i15` and `i16` still flake roughly 50% of the time. The remaining issue is in the `judgeInstructions` for both evals, which strictly require the agent's text response to include the literal domain string (`securityinsights-e2e-readonly-prod.apollo-df.dev`).

When prompted about "my domain", the agent retrieves the correct metrics and breakdowns, but frequently summarizes the findings as "for your domain" rather than repeating the full internal domain string. Whenever that string is omitted from the prose, the LLM judge fails the run.

Relaxing the `judgeInstructions` so repeating the raw domain string is optional won't reduce test rigor—the evals still strictly verify that the agent calls the correct tool and accurately reports the exact numerical findings (`8,627` visits, top 3 domains, transfer and user breakdowns). Alternatively, if explicitly confirming the queried tenant domain is desired UX, adding "Include the queried domain name" to the test prompt would also make both evals 100% stable.

This can be addressed here or later in another PR.

dylan-klein (approved):

PR #318, opened 2026-06-09
fix(eval): simplify judge instructions in i12

TusharKanekiDey:
Simplifies the judge instructions in `i12-check_security_insights.eval.js` to avoid penalizing the agent for omitting tool execution details in its response text, since the deterministic `expectedTools` assertion already verifies the tool call.


dylan-klein (approved):
LGTM. Tested locally and verified it resolves the flakiness in i12.

PR #319, opened 2026-06-09
fix(eval): update m05 target sub-OU to Sales

TusharKanekiDey:
Updates the target sub-OU path in the `m05` evaluation from the non-existent `/corp/sales` to the valid `/Sales` sub-OU that exists in the base mock database state. This prevents the agent from failing validation during OUs check.

### DIFF
--- a/test/evals/cases/mutation/m05-enable_security_insights_subou.eval.js
+++ b/test/evals/cases/mutation/m05-enable_security_insights_subou.eval.js
@@ -22,9 +22,13 @@ export default {
   forbiddenPatterns: [],
   requiredPatterns: [],
   prompt:
-    "Please enable Chrome Security Insights for our sub-organizational unit '/corp/sales' so we can stream its connectors.",
+    "Please enable Chrome Security Insights for our sub-organizational unit '/Sales' so we can stream its connectors.",
   goldenResponse:
-    'The agent should call the `security_insights` tool with `action: "enable"` and `targetOus: ["/corp/sales"]` specifically targeting the requested sub-OU path. It should confirm in plain language that the setup has been completed.',
+    'The agent should call the `security_insights` tool with `action: "enable"` and `targetOus: ["/Sales"]` ' +
+    'specifically targeting the requested sub-OU path. It should confirm in plain language that the setup ' +
+    'has been completed.',
   judgeInstructions:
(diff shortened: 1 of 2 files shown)


test/evals/cases/mutation/m05-enable_security_insights_subou.eval.js
```diff
@@ -22,9 +22,13 @@ export default {
   forbiddenPatterns: [],
   requiredPatterns: [],
   prompt:
-    "Please enable Chrome Security Insights for our sub-organizational unit '/corp/sales' so we can stream its connectors.",
+    "Please enable Chrome Security Insights for our sub-organizational unit '/Sales' so we can stream its connectors.",
```
dylan-klein, line 26: Add `scenario: 'healthy'` to this file (e.g., after `tags`). Without it, the test runs on the bare-minimum default state which lacks the `/Sales` OU. Correct agents will call `list_org_units`, see it's missing, and abort—causing the test to fail.

test/evals/cases/mutation/m05-enable_security_insights_subou.eval.js
```diff
@@ -22,9 +22,13 @@ export default {
...
   goldenResponse:
-    'The agent should call the `security_insights` tool with `action: "enable"` and `targetOus: ["/corp/sales"]` specifically targeting the requested sub-OU path. It should confirm in plain language that the setup has been completed.',
+    'The agent should call the `security_insights` tool with `action: "enable"` and `targetOus: ["/Sales"]` ' +
+    'specifically targeting the requested sub-OU path. It should confirm in plain language that the setup ' +
+    'has been completed.',
   judgeInstructions:
-    'Verify the agent indicator that it successfully enabled Security Insights for the requested sub-OU by calling `security_insights` with `action: "enable"` and `targetOus: ["/corp/sales"]`. Confirm it reported success in plain language.',
+    'Verify the agent indicator that it successfully enabled Security Insights for the requested sub-OU by ' +
```
dylan-klein: `judgeInstructions` seems like it needs to be adjusted to match PR #318 because the LLM judge only sees the final text response and cannot verify tool call details. `expectedTools` already handles the tool verification.

dylan-klein (changes requested):

TusharKanekiDey:
I have resolved the requested changes by: 1. Setting `scenario: 'healthy'` in the test case so that the `/Sales` organizational unit exists in the mock environment state. 2. Simplifying the `judgeInstructions` to avoid checking raw tool execution details in the final text response.

dylan-klein (approved):

PR #331, opened 2026-06-10
feat: implement security_insights_data tool and tests

sahil-madeka:
# Commit Summary: `704766b6114eb5d8a7df11add9fb2f0e05d2a5fa`

**Message**: `feat: implement security_insights_data tool and tests`
**Author**: Sahil Madeka 
**Date**: Wed Jun 10 12:51:06 2026 -0700

This commit implements the `security_insights_data` MCP tool to allow querying Chrome Enterprise Security Insights data (summaries and breakdowns of content transfers and URL visits), along with its registration and unit tests.

---

### 1. `tools/definitions/security-insights-data-tool.js` (NEW)
*   Defines and registers the `security_insights_data` MCP tool.
*   Supports four query actions:
    *   `queryContentTransfers` (calls `chromeManagementClient.queryContentTransfers`)
    *   `queryContentTransfersBreakdowns` (calls `chromeManagementClient.queryContentTransfersBreakdowns`)
    *   `queryUrlVisits` (calls `chromeManagementClient.queryUrlVisits`)
    *   `queryUrlVisitsBreakdowns` (calls `chromeManagementClient.queryUrlVisitsBreakdowns`)
*   Implements Zod input validation schema supporting parameters:
    *   `action` (Required, enum of the 4 actions above)
    *   `customerId` (Optional, defaults to `"my_customer"`)
    *   `filter` (Optional, AIP-160 filter string)
    *   `pageSize` (Optional, integer 1-1000, for breakdowns)
    *   `pageToken` (Optional, for breakdown pagination)
    *   `metric` (Optional, metric name for breakdowns)
    *   `breakdown` (Optional, dimension to break down by)
    *   `fixedTimeRange` (Optional, precomputed breakdown time range enum)
*   Uses `guardedToolCall` to handle authentication, retry logic, and automatic `customerId` resolution via `adminSdk` if omitted.
*   Formats the API JSON responses into human-readable text summaries and includes the raw structured data.

### 2. `tools/index.js`
*   Imports `registerSecurityInsightsDataTool`.
*   Registers the `security_insights_data` tool, gated by the `SECURITY_INSIGHTS_DATA_TOOL_ENABLED` feature flag.

### 3. `test/local/security_insights_data.test.js` (NEW)
*   Implements unit tests for the `security_insights_data` tool using Node's built-in `node:test` framework.
*   **Feature Flag Gating Tests**: Verifies the tool is registered only when `EXPERIMENT_SECURITY_INSIGHTS_DATA_TOOL_ENABLED` is set to `"true"`.
*   **Handler Logic Tests**:
    *   Mocks `ChromeManagementClient` query methods and verifies they are called with correct parameters.
    *   Verifies output formatting for both summary and breakdown actions.
    *   Verifies automatic customer ID resolution using `adminSdk` when `customerId` is omitted in the tool call.

## Manual test

Asked Gemini CLI to query content transfers for my domain, give me a breakdown of content transfers for my domain, asked "what urls are being visited in chrome in my domain". For the first case it called queryContentTransfers only and summarized its output, the second case it called queryContentTransfersBreakdowns only and summarized its output, and in the third case it called both queryUrlVisits and queryUrlVisits breakdowns and summarized the output of both using tables. 

In the next PR I will add a knowledge document to give more context about each of these APIs along with eval test cases


tools/definitions/security_insights_data_tool.js
```diff
@@ -0,0 +1,163 @@
...
+    },
+    guardedToolCall(
+      {
+        handler: async (
+          { action, customerId = 'my_customer', filter, pageSize, pageToken, metric, breakdown, fixedTimeRange },
+          { _requestInfo, authToken },
+        ) => {
+          logger.debug(`${TAGS.MCP} Calling 'security_insights_data' with action: ${action}, customerId: ${customerId}`)
+
+          const queryOptions = {}
```
dylan-klein, line 106: Optional: To make the tool more robust against "parameter pollution," we should sanitize the `queryOptions` object before passing it to the API client. Currently, all input parameters (like `pageSize`, `breakdown`, `fixedTimeRange`) are added to `queryOptions` and sent as query parameters to the Google API, regardless of the `action`. Because Google API gateways are typically strict about unknown query parameters, passing breakdown-specific parameters to a summary action (like `queryContentTransfers`) will likely result in a `400 Bad Request` error. We should conditionally add the breakdown-specific parameters to `queryOptions` only if the selected `action` ends with `'Breakdowns'`.

tools/definitions/security_insights_data_tool.js
```diff
@@ -0,0 +1,163 @@
...
+          return safeFormatResponse({
+            rawData: result,
+            toolName: 'security_insights_data',
+            formatFn: raw => {
+              let summaryText = `Security Insights action \`${action}\` completed.`
+              if (action === 'queryContentTransfers' || action === 'queryUrlVisits') {
+                const summaries = raw?.summaries || []
+                summaryText += ` Found ${summaries.length} summaries.`
+              } else {
+                const breakdowns = raw?.contentTransfersBreakdowns || raw?.urlVisitsBreakdowns || []
```
dylan-klein: Optional: Conflating the extraction of two different API response schemas using `||` fallback carries a maintenance risk. If the API response structure changes in the future, or if a developer incorrectly mocks the API response in a test, this fallback can silently select the wrong field or return an empty array instead of failing. This can result in misleading summary text (e.g., reporting "Found X breakdowns" using data from the wrong field) and makes debugging schema mismatches harder. It would be safer to explicitly map the expected response key based on the active `action` (e.g. using a ternary or conditional block to select either `contentTransfersBreakdowns` or `urlVisitsBreakdowns`).

dylan-klein (approved):

tools/definitions/security_insights_data_tool.js
sahil-madeka: done

tools/definitions/security_insights_data_tool.js
sahil-madeka: done

PR #333, opened 2026-06-10
feat(diagnose): proactively include Admin Console deep-links in environment health check findings

TusharKanekiDey:
This PR addresses [issue 522428294](https://b.corp.google.com/issues/522428294) by proactively providing Google Admin Console deep-links for Chrome Enterprise connectors, DLP rules, SEB apps, and Security Insights.

### Changes:
1. **Health Check Diagnostics (`diagnose_environment.js`)**: Updated `computeIssues` to directly include manual remediation links in the computed health issue messages.
2. **Capabilities Guide (`0-agent-capabilities.md`)**: Added a dedicated reference section `### 7. Google Admin Console Deep-Links for Remediation` documenting the exact settings page deep-links.
3. **Unit Tests (`diagnose_environment.test.js`)**: Added test case checking that diagnostic issues correctly contain the deep-links.
4. **Evals case (`i13-health_check_deep_links.eval.js`)**: Added a new evaluation case mapping to the `download-connector-missing` scenario to assert that the agent can surface the correct deep-links for manual configuration gaps.

### DIFF
--- a/tools/definitions/diagnose_environment.js
+++ b/tools/definitions/diagnose_environment.js
@@ -79,7 +79,10 @@ function computeIssues(data) {
   }
 
   for (const [key, connector] of Object.entries(data.connectors || {})) {
-    const name = CONNECTOR_DISPLAY_NAMES[key] || key
+    if (!Object.prototype.hasOwnProperty.call(CONNECTOR_DISPLAY_NAMES, key)) {
+      continue
+    }
+    const name = CONNECTOR_DISPLAY_NAMES[key]
     const page = CONNECTOR_LINK_MAPPING[key]
     const manualLink = page ? `https://admin.google.com/ac/chrome/settings/user/details/${page}` : null
     const actionSuffix = manualLink ? `. Update settings manually at ${manualLink}` : ''
--- a/tools/definitions/diagnose_environment.js
+++ b/tools/definitions/diagnose_environment.js
@@ -69,19 +69,32 @@ function computeIssues(data) {
     })
   }
 
+  const CONNECTOR_LINK_MAPPING = {
+    uploadAnalysis: 'file_attached',
+    downloadAnalysis: 'file_downloaded',
+    pasteAnalysis: 'bulk_text_entry',
+    printAnalysis: 'print_analysis_connector',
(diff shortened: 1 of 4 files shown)


tools/definitions/diagnose_environment.js
```diff
@@ -69,19 +69,35 @@ function computeIssues(data) {
...
+    downloadAnalysis: 'file_downloaded',
+    pasteAnalysis: 'bulk_text_entry',
+    printAnalysis: 'print_analysis_connector',
+    realtimeUrlCheck: 'realtime_url_check',
+    securityEventReporting: 'on_security_event',
+  }
+
   for (const [key, connector] of Object.entries(data.connectors || {})) {
-    const name = CONNECTOR_DISPLAY_NAMES[key] || key
+    if (!Object.prototype.hasOwnProperty.call(CONNECTOR_DISPLAY_NAMES, key)) {
```
dylan-klein, line 82: Optional: Silently skipping unknown connectors in the diagnosis loop might make it hard to detect bugs when new connectors are introduced. If a developer adds a new connector to `CONNECTOR_TYPES` but forgets to register it in `CONNECTOR_DISPLAY_NAMES` (in `lib/constants.js`), the connector will be silently ignored during diagnostics.

To mitigate this, consider one of the following:
1. **Unify the structures**: Define a single source of truth for connectors (e.g. in `lib/constants.js`) that groups the key, API type, display name, and remediation path, and use it to derive the other mappings.
2. **Add a unit test**: Export `CONNECTOR_TYPES` and `CONNECTOR_LINK_MAPPING` and add a test in `diagnose_environment.test.js` ensuring all keys in `CONNECTOR_TYPES` exist in `CONNECTOR_DISPLAY_NAMES` and `CONNECTOR_LINK_MAPPING`.
3. **Fallback behavior**: Keep the previous fallback behavior (using the raw key as a name) or log a warning when an unknown key is encountered, rather than silently skipping.

dylan-klein (approved):

tools/definitions/diagnose_environment.js
TusharKanekiDey: Will cut a bug for the above suggested changes and add it in later PR

PR #345, opened 2026-06-22
feat(auth): gate CLOUD_PLATFORM scope behind feature flag

sahil-madeka:


lib/util/credential/auth_login.js
```diff
@@ -64,6 +65,9 @@ export async function isTokenLocallyValid({
   cachePath = TokenCache.defaultPath(),
   scopes = [],
 } = {}) {
+  if (!featureFlags.isEnabled(FLAGS.CLOUD_PLATFORM_SCOPE_ENABLED)) {
```
dylan-klein: Currently, the static MCP tool description for `cep_auth` (`tools/definitions/auth.js#L164`) is still built using unfiltered `Object.values(SCOPES)`. This means connected AI agents are told the server requests the `Google Cloud Platform` permission even when this flag is disabled and filtered out at runtime.

Optional: Fixing this scattered filtering across a 5th file highlights a refactor opportunity: instead of repeating `if (!featureFlags.isEnabled...)` across `auth_login.js`, `cli_commands.js`, `oauth_flow.js`, `mcp-server.js`, and `auth.js`, let's export a single `getActiveScopes()` helper in `lib/constants.js` (or `feature_flags.js`) as the canonical source of truth:
```js
/**
 * Returns the active OAuth scope array based on current feature flag settings.
 * @returns {string[]} Array of active scope URLs.
 */
export function getActiveScopes() {
  const scopes = Object.values(SCOPES)
  if (!featureFlags.isEnabled(FLAGS.CLOUD_PLATFORM_SCOPE_ENABLED)) {
    return scopes.filter(s => s !== SCOPES.CLOUD_PLATFORM)
  }
  return scopes
}
```
Then every entry point (including `scopeSummary` in `auth.js` and default parameters in `guardedToolCall`) can simply call `getActiveScopes()` instead of `Object.values(SCOPES)`.

dylan-klein (changes requested):

lib/util/credential/auth_login.js
sahil-madeka: done

dylan-klein (approved):

PR #346, opened 2026-06-22
feat(Crm): Add search_organizations tool to MCP server

sahil-madeka:
This PR adds a new tool called search_organizations which finds the GCP org for a given customer. This is then stored in the session state for tools that will require a GCP organization. If no organization is found then that member of sessionState is left empty. The output of this tool is the organization ID and name. If it is empty MCP clients are instructed to suggest to the user to create a GCP organization ID. 

## Manual test

I created a custom OAuth client as the default one does not have the cloud platform scope added yet. With the custom client I was able to get the call to work with Gemini CLI by using the prompt "can you search my organization using cep-dev?"
SearchOrganizationResult

### DIFF
--- a/test/integration/tools/search_organizations.test.js
+++ b/test/integration/tools/search_organizations.test.js
@@ -48,9 +48,9 @@ describe('Search Organizations Tool Integration', () => {
     assert.ok(text.includes('Test Org'), `Output text: ${text}`)
     assert.ok(text.includes('123456789'), `Output text: ${text}`)
 
-    // Verify raw details in structured content
-    assert.strictEqual(details.organizations[0].displayName, 'Test Org')
-    assert.strictEqual(details.organizations[0].name, 'organizations/123456789')
+    // Verify structured content details
+    assert.strictEqual(details.organizationId, '123456789')
+    assert.strictEqual(details.organizationName, 'organizations/123456789')
 
     // Verify sessionState was updated
     assert.strictEqual(sessionState.organizationId, '123456789')
--- a/test/integration/tools/search_organizations.test.js
+++ b/test/integration/tools/search_organizations.test.js
@@ -56,22 +56,4 @@ describe('Search Organizations Tool Integration', () => {
     assert.strictEqual(sessionState.organizationId, '123456789')
     assert.strictEqual(sessionState.organizationName, 'organizations/123456789')
   })
-
-  test('When search_organizations is called without customerId, then it auto-resolves and succeeds', async () => {
-    const { client, sessionState } = harness
-
-    // Reset session state to verify auto-resolution writes to it again
-    sessionState.organizationId = null
-    sessionState.organizationName = null
-
-    const result = await client.callTool({
-      name: 'search_organizations',
-      arguments: {}, // Omit customerId
-    })
-
-    const { text } = parseToolOutput(result)
-
-    assert.ok(text.includes('Associated GCP Organization Found'), `Output text: ${text}`)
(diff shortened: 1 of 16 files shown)


tools/definitions/search_organizations.js
```diff
@@ -0,0 +1,108 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * @file Tool definition for searching GCP organizations.
```
dylan-klein, line 18: You will need to update the PR title to follow Conventional Commits (e.g., `feat(tools): add search_organizations tool and CRM integration`). Currently, `Crm tool integration` will fail commit validation when squash-merging. 

For context on what mandates `feat()`: our CI check in [`.github/workflows/commitlint.yml`](https://github.com/google/chrome-enterprise-premium-mcp/blob/main/.github/workflows/commitlint.yml) executes [`commitlint.config.js`](https://github.com/google/chrome-enterprise-premium-mcp/blob/main/commitlint.config.js), which imports `@commitlint/config-conventional`. That npm preset defines the official type grammar (`feat`, `fix`, `chore`, `docs`, etc.), which is also documented in our repo steering guide ([`AGENTS.md`](https://github.com/google/chrome-enterprise-premium-mcp/blob/main/AGENTS.md#workflow)).

tools/definitions/search_organizations.js
```diff
@@ -0,0 +1,108 @@
...
+    'search_organizations',
+    {
+      description:
+        'Searches for the GCP organization associated with the Workspace customer. ' +
+        'Resolves and caches the Organization ID in the session for subsequent Context-Aware Access (CAA) operations.',
+      inputSchema: z.object({
+        customerId: z
+          .string()
+          .optional()
+          .describe('The Workspace customer ID (e.g. C012345). Defaults to "my_customer".'),
```
dylan-klein: Are we sure Cloud Resource Manager also supports `"my_customer"` as a tenant alias? Unlike Admin SDK Directory APIs, GCP IAM services usually expect literal directory customer IDs (e.g. `C012345`). If an LLM reads `.describe('... Defaults to "my_customer"')` and explicitly passes `customerId: "my_customer"`, `guardedToolCall` won't auto-resolve it. Might be worth double checking if CRM interprets that string. We could potentially intercept `'my_customer'` if not.

dylan-klein (changes requested):

tools/definitions/search_organizations.js
sahil-madeka: fixed the description

tools/definitions/search_organizations.js
sahil-madeka: done

test/integration/tools/search_organizations.test.js
```diff
@@ -0,0 +1,77 @@
+/*
```
dylan-klein, line 1: Please add evals at some point to ensure tool description is being interpreted correctly. Personally, I see some merit in adding evals in same PRs that any natural language instruction is added, but it can be added in later PR if preferred.

sahil-madeka: I added evals for the two main scenarios

dylan-klein (approved):

PR #347, opened 2026-06-22
test: hermetically sanitize custom OAuth client env vars in runners

dylan-klein:
Stops developers from seeing false test failures caused by local environment variables or `.env` files on their machine.

### Motivation
- If an engineer has `CEP_OAUTH_CLIENT_ID` and `CEP_OAUTH_CLIENT_SECRET` set in their terminal or in a local `.env` file, test server subprocesses inherit those credentials.
- This causes the server to boot in custom OAuth mode and unexpectedly register the `check_and_enable_cep_api` tool, failing stdio tool-listing test assertions.
- Blanking out these variables before running tests guarantees hermetic, consistent test results for everyone regardless of local machine configuration.

### Main Changes
- `test/run-unit.js`: Blank out custom OAuth client environment variables before running unit tests.
- `test/run-integration.js`: Blank out custom OAuth client variables before running fake integration tests (preserved for live `real` runs so engineers can still test against their BYO OAuth client projects).
- `test/integration/server/mcp-server.test.js`: Restored explicit blanking inside the `before()` hook so direct IDE test runs (`node --test ...`) remain isolated.

### Verification
- Simulated leaking `CEP_OAUTH_CLIENT_ID=leak-test` into `npm test`; confirmed all 39 integration tests pass cleanly.
- Ran full `npm run presubmit` suite (all checks pass).

### DIFF
--- a/test/run-integration.js
+++ b/test/run-integration.js
@@ -34,7 +34,7 @@ limitations under the License.
 import { join, resolve, dirname } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import { execFileSync } from 'node:child_process'
-import { findTestFiles } from './run-utils.js'
+import { findTestFiles, sanitizeOauthClientEnv } from './run-utils.js'
 import { setupSyntheticTokenCache } from './helpers/synthetic_token_cache.js'
 
 const __dirname = dirname(fileURLToPath(import.meta.url))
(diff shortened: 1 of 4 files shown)


test/run-integration.js
```diff
@@ -55,6 +55,10 @@ if (backend !== 'fake' && backend !== 'real') {
 }
 
 process.env.CEP_BACKEND = backend
+if (backend !== 'real') {
```
sahil-madeka, line 58: should we set these back to their originally values before the test was run when its finished running?

dylan-klein: Since `run-integration.js` is a standalone CLI entrypoint executed in its own OS process (`node test/run-integration.js`), the Node process exits immediately after `execFileSync` finishes. The OS automatically destroys the entire process memory space and `process.env` table on exit, leaving the parent shell environment untouched.

(Note that `setupSyntheticTokenCache()` redirects `process.env.HOME` without restoring it on exit for the exact same reason).

sahil-madeka (approved):

test/run-integration.js
dylan-klein: Also extracted this sanitization logic into a centralized helper (`sanitizeOauthClientEnv` in `test/run-utils.js`) with explicit JSDoc documentation explaining why blanking is required. Re-requesting review.

PR #363, opened 2026-06-30
ci: add PR eval presubmit check with 80% pass rate threshold

dylan-klein:
### Motivation
Currently, agent evals only run on a daily scheduled cron job. When a contributor adds or modifies an eval case, there is no automated presubmit verification on the pull request, creating a risk that broken or flaky evals land in `main`.

### Main Changes
* Added `.github/workflows/pr-evals.yml` to automatically trigger when files under `test/evals/cases/` change in a Pull Request targeting `main`.
* Uses `git diff --name-only` against the PR base branch to dynamically identify only the specific `.eval.js` files modified or added by the PR.
* Executes 5 evaluation runs (`--runs 5`) per changed eval case and parses the generated `pr-evals.json` report to enforce an **80% pass rate threshold** (at least 4 out of 5 runs passing).

### Testing
Verified live in GitHub Actions using a stacked test branch containing a deliberately broken eval case (`i19` with an impossible required string pattern). The workflow correctly identified the changed eval file, executed 5 runs, detected the 0% pass rate, and failed the status check with exit code 1 as required:
* Live verification run: https://github.com/google/chrome-enterprise-premium-mcp/actions/runs/28467964537

### Design Decisions
* **Fork vs. Internal PR Handling**: GitHub Actions strips repository secrets (like `GEMINI_API_KEY`) from `pull_request` workflows originating from external forks. The workflow checks for secret availability: if `GEMINI_API_KEY` is present (internal PRs), it runs the full 5-run agent+judge suite; if missing (external forks), it automatically falls back to `--dry-run` mode to validate config syntax and deterministic checks without failing on auth.
* **Inline Threshold Parsing**: Because `test/evals/run.js` natively exits with code 1 if *any* single run fails, we execute it with `|| true` and use a short inline Node script to verify `e.passRate < 80`, allowing minor transient LLM flakiness without failing the CI check.

TAG=agy
CONV=a699777c-978b-40ae-a24d-9847b07c0ae4

### DIFF
--- a/.github/workflows/pr-evals.yml
+++ b/.github/workflows/pr-evals.yml
@@ -1,12 +1,6 @@
 name: PR Eval Check
 
 on:
-  push:
-    branches:
-      - 'ci/**'
-    paths:
-      - 'test/evals/cases/**/*.eval.js'
-      - '.github/workflows/pr-evals.yml'
   pull_request:
     branches: [main]
     paths:
--- a/.github/workflows/pr-evals.yml
+++ b/.github/workflows/pr-evals.yml
@@ -1,6 +1,12 @@
 name: PR Eval Check
 
 on:
+  push:
+    branches:
+      - 'ci/**'
+    paths:
+      - 'test/evals/cases/**/*.eval.js'
+      - '.github/workflows/pr-evals.yml'
   pull_request:
     branches: [main]
     paths:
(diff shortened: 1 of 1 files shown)


TusharKanekiDey:
Great feature !
Thanks for adding this.

TusharKanekiDey (approved):

PR #365, opened 2026-06-30
fix(eval): resolve connector validation issues in healthy base scenario (b/527496693)

TusharKanekiDey:
Sets delayDeliveryUntilVerdict to true for the bulk text entry connector and adds the missing core reporting events (unscannedFileEvent and interstitialEvent) to the default base state in base-state.js. This ensures that the healthy scenario does not report false-positive High severity issues, preventing failures in pr01 and pr17.


sahil-madeka:
Can we add a testing section in the PR description with the command used to test these and also the percent success rate?

dylan-klein (approved):

PR #368, opened 2026-07-06
docs: update knowledge article to cite security_insights_data tool 

sahil-madeka:
This PR adds a knowledge document for the Query insights data API. The public doc is currently ACLd so we need to provide a body describing the APIs


brettmeehan-google:
Since the Security Insights v1 docs are now public should we link those too?

sahil-madeka:
> Since the Security Insights v1 docs are now public should we link those too?

i will make a separate PR to call the v1 endpoints and add references to public docs in that PR

brettmeehan-google (approved):

PR #375, opened 2026-07-08
fix: bypass OAuth token check when service account is configured

shantand1234:
If a service account is configured via GOOGLE_APPLICATION_CREDENTIALS, the local OAuth pre-flight verification in guardedToolCall should be bypassed. Otherwise, users are incorrectly blocked with a 'Sign-in is needed' error if they have an expired or missing OAuth access token, even though the service account credentials are valid and would be successfully utilized by the API clients.

Also refactors auth_messages.js to appropriately report and display Service Account status in the console banner UI.

### DIFF
--- a/lib/util/auth_messages.js
+++ b/lib/util/auth_messages.js
@@ -44,6 +44,10 @@ function labelFor(source) {
  * @returns {string} The banner field text.
  */
 export function buildScopesField(probe, requiredScopes) {
+  if (process.env.GOOGLE_APPLICATION_CREDENTIALS) {
+    return `🟢 Service Account configured`
+  }
+
   if (probe.scopesKnown && probe.missingScopes.length > 0) {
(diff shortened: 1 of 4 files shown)


lib/util/auth_messages.js
```diff
@@ -44,6 +44,10 @@ function labelFor(source) {
  * @returns {string} The banner field text.
  */
 export function buildScopesField(probe, requiredScopes) {
+  if (process.env.GOOGLE_APPLICATION_CREDENTIALS) {
+    return `🟢 Service Account configured`
```
dylan-klein, line 48: nit: Use single quotes `'🟢 Service Account configured'` instead of backticks since there is no string interpolation (`${...}`) and surrounding static returns in this function use single quotes.

dylan-klein (approved):

PR #389, opened 2026-07-15
feat(auth): proactive regime-aware authentication error remediation and DWD 400 interception

dylan-klein:
### Motivation
When the Chrome Enterprise Premium MCP server (`chrome-enterprise-premium-mcp`) is accessed via Service Account credentials (such as when connected through Pocket CEP or remote HTTP/SSE callers), several authentication and delegation friction points occurred:
1. When a direct Service Account (without `CEP_IMPERSONATE_SUBJECT` / user impersonation) called directory or policy tools requiring human admin delegation (`requiresDelegation: true`), Google returned `400 Bad Request / Invalid Customer Id` instead of `403 Forbidden`. Previously, this raw `400` error was passed through to the agent, causing false alarms about Customer ID validity rather than guiding the user to configure Domain-Wide Delegation.
2. In direct/delegated Service Account modes and HTTP/Bearer modes, the server occasionally suggested interactive CLI commands (`gcloud auth login` / `cep_auth`), which cannot be executed inside headless or remote client environments.
3. During local `npm run presubmit` smoke testing, `smoke-test.js` hardcoded port `3000`/`3099`, causing port conflict crashes (`EADDRINUSE`) whenever a developer's Next.js dev server (`pocket-cep`) was running in the background.

### Main Changes
* **Principal Identity Extraction & Interception (`wrapper.js`)**:
  * Implemented `getAuthenticatedPrincipalInfo` to inspect `GOOGLE_APPLICATION_CREDENTIALS`, `CEP_IMPERSONATE_SUBJECT`, and inbound JWT token subjects (`client_email`, `iss`, `sub`).
  * Updated the `requiresDelegation` pre-flight check and `isAuthError` interceptors so when `requiresDelegation === true` and `isServiceAccountPrincipal && !isImpersonating`, **both `status === 403` and `status === 400` / `Invalid Customer Id` errors** are caught. Instead of surfacing a raw `400` Customer ID error, the server returns an actionable Domain-Wide Delegation setup recommendation pointing to the Admin Console (`admin.google.com/ac/owl/domainwidedelegation`).
* **Mutually Exclusive Error Remediation (`auth.js`, `auth-error.js`, `wrapper.js`)**:
  * Strictly separated DWD delegation failures from Workspace Admin Console privilege/role assignment failures (`TOOL_PRIVILEGES_MAP[toolName]`), displaying exact required admin privileges ("User management", "Data Loss Prevention (DLP)", etc.).
  * Suppressed all interactive local terminal login instructions (`gcloud auth login` / `npx ... auth login`) whenever Service Account or Bearer authentication modes are active.
* **Dynamic OS-Assigned Smoke Ports (`smoke-test.js`)**:
  * Replaced hardcoded ports (`3000` / `3099`) with `PORT: 0` (letting the OS assign a random free port) across all three HTTP/SSE test phases (`runToolTest`, `runPromptTest`, and `runResourceTest`), eliminating `EADDRINUSE` failures during `npm run presubmit`.

### Verification & Testing
* **Unit Tests (`auth.test.js`, L280-322)**: Added unit tests verifying `guardedToolCall` delegation guards intercepting `status === 400 Invalid Customer Id` when calling `requiresDelegation: true` tools without impersonation (`!isImpersonating`).
* **Full Presubmit Pass**: Ran `npm run presubmit` (`npm run test:unit`, `npm run test:integration:fake`, and `npm run test:smoke`). 100% of all ~589 unit, integration, and smoke tests passed cleanly without port conflicts.

BUG=533506635
CONV=6b6f341c-df70-4e0b-b879-be57f38b11ea
TAG=agy


lib/util/auth_mode.js
```diff
@@ -0,0 +1,65 @@
...
+limitations under the License.
+*/
+
+/**
+ * @file Helper utilities for normalizing and checking CEP_AUTH_MODE configurations.
+ */
+
+/**
+ * Returns the normalized auth mode: 'dynamic' | 'bearer-only' | 'service-account-only'.
+ * @returns {'dynamic' | 'bearer-only' | 'service-account-only'} The active auth mode.
```
shantand1234, line 23: the service account mode here is independent of pocket cep right?
When running pocket cep I assume that the pocket cep will be running in service account mode and the mcp server will be working in bearer modec

shantand1234 (approved):

lib/util/auth_mode.js
dylan-klein: Correct. To confirm PocketCEP's auth was being used, I tested starting up CEP MCP with `CEP_AUTH_MODE=bearer-only` and PocketCEP with `AUTH_MODE=service_account`: this worked. I then switched CEP MCP to `CEP_AUTH_MODE=service-acount-only` and PocketCEP tool calls no longer worked, because CEP MCP was never given its own credentials to use.

PR #392, opened 2026-07-20
docs(acm): update list_caa_access_levels tool description to check existing levels before creation

sahil-madeka:
This PR updates the description of list access levels to direct agentic clients to call it first to see if an access level meeting the users conditions already exists before creating one 

## Manual Testing

Screenshot 2026-07-20 9 04 25 AM

### DIFF
--- a/tools/definitions/list_caa_access_levels.js
+++ b/tools/definitions/list_caa_access_levels.js
@@ -81,6 +81,8 @@ export function registerListCaaAccessLevelsTool(server, options, sessionState) {
       description: `Lists Context-Aware Access (CAA) access levels for an organization.
 Resolves the Access Policy for the specified (or cached) Organization ID, then lists all defined access levels within that policy.
 
+When a user requests to create a new access level, agents should first invoke 'list_caa_access_levels' to check if an access level matching the user's described conditions already exists for the organization. If a matching access level is found, present that access level to the user and ask if they still wish to create a new one before proceeding with creation.
+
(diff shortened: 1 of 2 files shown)


tools/definitions/list_caa_access_levels.js
```diff
@@ -81,6 +81,8 @@ export function registerListCaaAccessLevelsTool(server, options, sessionState) {
       description: `Lists Context-Aware Access (CAA) access levels for an organization.
 Resolves the Access Policy for the specified (or cached) Organization ID, then lists all defined access levels within that policy.
 
+When a user requests to create a new access level, agents should first invoke 'list_caa_access_levels' to check if an access level matching the user's described conditions already exists for the organization. If a matching access level is found, present that access level to the user and ask if they still wish to create a new one before proceeding with creation.
```
dylan-klein, line 84: Why add this instruction to `list_caa_access_levels`? It might make more sense to have it in `create_caa_access_level` instead (or in both), since we want to guard the use of the create tool. If it's only in list, an agent looking to create might miss it.

dylan-klein (commented):
Consider adding an eval to verify that the agent actually follows this instruction (i.e., it checks for existing levels and prompts the user if a match is found).

tools/definitions/list_caa_access_levels.js
sahil-madeka: I believe that agents usually read all the tools' description when the MCP server is loaded/reloaded, is that not always the case?

I updated creates description regardless so its in both places now

dylan-klein (approved):

PR #400, opened 2026-08-05
feat(mcp): add READ_ONLY startup flag to disable mutating tools

dylan-klein:
### Motivation
Introduce a secure read-only mode to allow customers to safely try the Chrome Enterprise Premium (CEP) MCP server and perform audits or diagnostics without the risk of accidental policy modifications or resource creation in their GCP/Chrome environments.

### Main Changes
*   **Feature Flag:** Added `READ_ONLY` flag to `lib/util/feature_flags.js` (enabled via the `EXPERIMENT_READ_ONLY` environment variable).
*   **Capability-Based Tagging:** Updated `guardedToolCall` in `tools/utils/wrapper.js` to require an `isMutating` boolean parameter. This ensures all wrapped tools explicitly declare whether they mutate state. Failure to declare this throws a developer error at startup.
*   **Centralized Enforcer Proxy:** Implemented a Proxy in `tools/index.js` that wraps the server.
    *   It validates that all registered tools (except `cep_auth`) declare `isMutating`.
    *   If `READ_ONLY` is active and `isMutating` is true, it prefixes the tool description and replaces the handler with a failing one.
*   **Hybrid Tool:** Updated `check_and_enable_cep_api` to reject `enable: true` calls with an error, while keeping read-only checks functional.
*   **Server Instructions:** Appends a status block to the server instructions to inform the agent that it is in read-only mode.
*   **Metadata Migration:** Migrated all 30+ tools and tests to explicitly declare their `isMutating` status.

### Design Decisions
*   **Decentralized Tagging with Fail-Fast Validation:** Decided against a hardcoded list of mutating tools in `tools/index.js` to prevent drift. Instead, tools self-declare `isMutating`. The registration proxy enforces this at startup, making it impossible for a developer to register a new tool without declaring its capability.
*   **Description Prefixing:** Prefixed descriptions instead of hiding tools entirely. This allows the agent to remain aware of the tools for planning purposes while knowing they cannot be executed.

TAG=agy
CONV=2de5746f-a0c6-425e-8332-30486f538151


TusharKanekiDey (approved):
I feel it can be broken down into smaller cl.
But it works this way as well

